PR #135, opened 2023-09-18
Fix resource plugin auto-setup

prolativ:
* Expose plugiDownloadURL from provider schemas to plugin.json to make community packages automatically installable
* Avoid clashes on plugin.json resources
* Enable building language plugin for different platforms


lbialy (changes requested):
add your test to CI setup if possible

lbialy (approved):

PR #137, opened 2023-09-20
Allow val defined as a const to be used as NES

Kordyjan:
This change allows references to values defined as non-empty constants to be used as NonEmptyStrings.
Works when the `-Yretain-trees` flag is enabled.

It can be taken further by applying the checks recursively, e.g., we can check make things like 

```scala
val x = s"abc${"xyz"}"
val nes: NonEmptyString = x
```

or 

```scala
val x = "ABC"
val y = "def"
val nes: NonEmptyString = s"$x$y"
```

work.

### DIFF
--- a/core/src/main/scala/besom/util/NonEmptyString.scala
+++ b/core/src/main/scala/besom/util/NonEmptyString.scala
@@ -51,9 +51,23 @@ object NonEmptyString:
             else if s.isBlank then report.errorAndAbort("Blank string is not allowed here!")
             else str.asExprOf[NonEmptyString]
           case None =>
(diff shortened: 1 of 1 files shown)


lbialy:
@prolativ can you please add -YretainTrees to default setup of Scala-cli launcher, templates and a test?

prolativ:
@lbialy is this something we need for our initial release? I would opt for holding this a bit to do some experiments and see how this pattern might influence users' code

lbialy:
Ok, sure.

lbialy:
This is subsumed by #375. Thank you @Kordyjan for all the help with this!

PR #153, opened 2023-09-22
Use compiler plugin in templates

### LINKED ISSUE #146: Find a way to detect wrong string interpolation
It would be helpful to the user if we could detect at compile time this subtle problem:
```
sourceArn = s"${api.executionArn}/*"
```

instead of runtime error:
```
Diagnostics:
  aws:lambda:Permission (feedLambdaPermission):
    error: aws:lambda/permission:Permission resource 'feedLambdaPermission' has a problem: "source_arn" (besom.internal.Output@6da60311/*) is an invalid ARN: arn: invalid prefix. Examine values at 'feedLambdaPermission.sourceArn'.
````

And recommend user to use our interpolator:
```
sourceArn = p"${api.executionArn}/*"
```

This is a very common problem in all other SDKs, so this is a nice-to-have. 
This might not be possible thou.

pawelprazak:
Fix #146

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .metals
 .vscode
 *.jar
+*.semanticdb
 *.iml
 target/
 .DS_Store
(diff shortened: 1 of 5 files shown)


lbialy (approved):

lbialy:
Is this also covered in your todays scope @pawelprazak?

PR #160, opened 2023-09-22
replacing codec macros with inlines

pawelsadlo:
If there is no need for custom errors it can be simplified.


prolativ:
LGTM in general. However I would hold merging that until we finish the beta release. Also we should see if/how this change influences compilation times from different projects from codegen, especially the big ones

lbialy:
I have verified this and macros are actually faster than raw inlines.

PR #172, opened 2023-09-26
Prepare templates before release

pawelprazak:
- add a GHA workflow
- add smoke tests using pulumi preview
- temporarily revert syntax to work with the released 0.0.1-beta version

Depends on #176


.github/workflows/templates.yml
```diff
@@ -0,0 +1,37 @@
+on:
+  push:
+    branches: [ 'main', 'develop' ]
+    paths:
+      - 'template'
```
prolativ, line 5:
```suggestion
      - 'templates'
```

.github/workflows/templates.yml
```diff
@@ -0,0 +1,37 @@
...
+  pull_request:
+    branches: [ 'main', 'develop' ]
+    paths:
+      - 'template'
```
prolativ, line 9:
```suggestion
      - 'templates'
```

.github/workflows/templates.yml
lbialy, line 5: templates

.github/workflows/templates.yml
lbialy, line 9: templates

templates/kubernetes/Main.scala
```diff
@@ -1,12 +1,12 @@
...
-import besom.api.kubernetes.apps.v1.{Deployment, DeploymentArgs}
+import besom.api.kubernetes.apps.v1.{deployment, DeploymentArgs}
 import besom.api.kubernetes.core.v1.inputs.{ContainerArgs, ContainerPortArgs, PodSpecArgs, PodTemplateSpecArgs}
 import besom.api.kubernetes.meta.v1.inputs.{LabelSelectorArgs, ObjectMetaArgs}
 import besom.api.kubernetes.apps.v1.inputs.DeploymentSpecArgs
 
 @main def main = Pulumi.run {
   val appLabels = Map("app" -> "nginx")
-  for nginxDeployment <- Deployment(
+  for nginxDeployment <- deployment(
```
lbialy, line 9: wait, what?

lbialy (changes requested):
we need this to be updated to 0.1.0 once we go live

templates/kubernetes/Main.scala
pawelprazak: beta thawa was released has old syntax as mentioned on slack, will update to 0.2 in next step

.github/workflows/templates.yml
pawelprazak: thank you! :)

pawelprazak:
> we need this to be updated to 0.1.0 once we go live

yes, I agree, but I'd prefer to merge this and do 0.1.0 update in next PR

pawelprazak:
Merged #176 here to resolve cyclic dependency

lbialy (approved):

PR #176, opened 2023-09-27
Add aws template

pawelprazak:
Depends on #172 
Is required for: #175


lbialy (approved):
It's dumb. It should be using at least 0.0.2-beta.

pawelprazak:
merged into #172

Issue #138, opened 2023-09-21
Inconvenient `NonEmptyString`


pawelprazak:
Currently, the API forces to explicitly type string `val`'s, e.g.:
```scala
val bucketName: NonEmptyString = "myResource"
```

It would be nice to find a way to solve this without sacrificing the type-safety.

lbialy:
Solved by #137

pawelprazak:
Using `"enable-" + api.replace(".", "-"),` should behave like `s"enable-${ api.replace(".", "-") }"`, got:

```
Only constant strings or string interpolations are allowed here, use NonEmptyString.apply instead!
```

lbialy:
Hopefully closed by #375!

lbialy:
Reopen if any other problems are found.

PR #186, opened 2023-09-28
Add aws-s3-folder example

pawelprazak:
add the first example and scaffolding


examples/aws-s3-folder/Main.scala
```diff
@@ -0,0 +1,88 @@
+import besom.*
+import besom.api.aws.*
+import besom.api.aws.s3.BucketObject
```
lbialy, line 3: Could we use the prefix notation like Pulumi consistently? I actually like it as it's very readable in Pulumi context, more than bare imported classes.

In this file I'd either import aws and use full aws.s3.BucketObject or import aws.s3 and use s3.BucketObject.

lbialy: You are actually using the s3.* syntax. Can you just use s3.BucketObject in the return type in the operation that maps files to BOs?

examples/aws-s3-folder/Main.scala
```diff
@@ -0,0 +1,88 @@
...
+      CustomResourceOptions(
+        dependsOn = siteBucketPublicAccessBlock.map(List(_))
+      )
+    )
+  )
+
+  // For each file in the directory, create an S3 object stored in `siteBucket`
+  val uploads: Array[Output[BucketObject]] = File(siteDir).listFiles().map { file =>
+    s3.BucketObject(
+      NonEmptyString(file.getName).get, // FIXME: https://github.com/VirtusLab/besom/issues/138
```
lbialy, line 67: That won't get fixed by #138. `file.getName` is a completely dynamic string value coming from java.io APIs. Functional approach would be to patmat that and handle error case by failing the Output (or skipping the file, it's up to the user). I'm actually not sure if we have an Output.fail. The discussion on error handling APIs didn't occur yet. Throwing works because we catch but it's not pure.

integration-tests/src/test/resources/logger-test/run.scala
```diff
@@ -0,0 +1,12 @@
+//> using scala 3.3.1
```
lbialy, line 1: Should these files be in this PR?

lbialy (changes requested):
Some questions

.gitignore
```diff
@@ -20,7 +20,6 @@ scoverage-reports/
 # generated sources, run `scala-cli run ./scripts -M proto -- all`
```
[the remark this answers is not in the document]
pawelprazak: this line was already there, not sure what is the proposition here

examples/aws-s3-folder/Main.scala
pawelprazak: sure, will do :)

integration-tests/src/test/resources/logger-test/run.scala
pawelprazak: definitely not, thank you

pawelprazak:
OK, made the requested changes, just waiting for the lang plugin fix to pass the CI

lbialy (approved):

lbialy:
something is no yes in Examples smoke tests

pawelprazak:
Thanks, fixed :)

PR #188, opened 2023-09-28
Fix packaging of the language plugin to correct use the specified OS and architecture

prolativ:


pawelprazak (approved):
LGTM, I've tested locally, it fixes the problem, thank you!

PR #206, opened 2023-10-02
Add aws-secrets-manager example

pawelprazak:


examples/aws-s3-folder/Main.scala
```diff
@@ -61,7 +61,7 @@ val siteDir = "www"
   )
 
   // For each file in the directory, create an S3 object stored in `siteBucket`
-  val uploads: Array[Output[s3.BucketObject]] = File(siteDir).listFiles().map { file =>
+  val uploads: Output[List[s3.BucketObject]] = File(siteDir).listFiles().map { file =>
```
KacperFKorban, line 64: nit: if it's an example then maybe we should use traverse, since `.map(...).sequence == .traverse(...)`?

examples/aws-s3-folder/project.scala
```diff
@@ -1,4 +1,6 @@
 //> using scala "3.3.1"
 //> using plugin "org.virtuslab::besom-compiler-plugin:0.1.0"
 //> using dep "org.virtuslab::besom-core:0.1.0"
-//> using dep "org.virtuslab::besom-aws:6.2.1-core.0.1.0"
\ No newline at end of file
+//> using dep "org.virtuslab::besom-aws:6.2.1-core.0.1.0"
+//> using options -Werror -Wunused:all -Wvalue-discard -Wnonunit-statement
+//> using options -Werror -Wunused:all -Wvalue-discard -Wnonunit-statement
```
KacperFKorban, line 6: nit: Unnecessary line?

examples/aws-secrets-manager/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+6. From there, feel free to experiment. Simply making edits and running pulumi up will incrementally update your infrastructure.
+
+7. To clean up resources, destroy your stack and remove it:
+
+    ```bash
+    pulumi destroy
+    ```
+    ```bash
+    pulumi stack rm
```
KacperFKorban, line 56: nit: could add the the stack name as well, to be more consistent.

KacperFKorban (approved):

examples/aws-secrets-manager/README.md
pawelprazak: I thought about this, but this way the user has to confirm the deletion, so I thought this would be safer, WDYT?

examples/aws-s3-folder/project.scala
pawelprazak: good catch, thanks! :)

examples/aws-secrets-manager/README.md
KacperFKorban: Makes sense, though I think that `pulumi stack rm stack-name` also requires confirmation.
image

pawelprazak: OK, I didn't recall correctly, explicit stack name does not disable the confirmation, will adjust, thank you! :)

examples/aws-s3-folder/Main.scala
```diff
@@ -61,29 +61,32 @@ val siteDir = "www"
   )
 
   // For each file in the directory, create an S3 object stored in `siteBucket`
-  val uploads: Array[Output[s3.BucketObject]] = File(siteDir).listFiles().map { file =>
+  val uploads: Output[List[s3.BucketObject]] = File(siteDir).listFiles().toList.traverse { file =>
```
pawelprazak, line 64: @lbialy you should be happier with those changes, @KacperFKorban helped me to make it cleaner :)

Issue #209, opened 2023-10-03
Add  missing `StackReference`


pawelprazak:
We are missing [Stack References](https://www.pulumi.com/learn/building-with-pulumi/stack-references/), also:
-  make sure to properly handle [organizations in Stack name](https://www.pulumi.com/docs/concepts/stack/#stackreferences) - an opaque type with extensions, like URI
- make sure we handle the new direct accessor - https://www.pulumi.com/docs/concepts/stack/#reading-outputs-from-stack-references

Might be related to #43

Issue #204, opened 2023-10-02
Improve `Config` ergonomy


pawelprazak:
Currently we are missing all `.require*` methods and the `.get*` methods return `Output[A]` and this results in silently ignoring missing keys with an empty output.

We might want to change the current behaviour to:
- `.get*` returning `Output[Option[A]]`
- `.require*` returning `Output[A]` with an error

The error content should be identical to the [reference implementation](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/config.ts#L524):
```typescript
/**
 * ConfigMissingError is used when a configuration value is completely missing.
 */
class ConfigMissingError extends RunError {
    constructor(public key: string, public secret: boolean) {
        super(
            `Missing required configuration variable '${key}'\n` +
                `\tplease set a value using the command \`pulumi config set${
                    secret ? " --secret " : " "
                }${key} <value>\``,
        );
    }
}
```

Also, the namespaced syntax should not require `map` and use lifting instead:
`Config("aws").map(_.get("region"))` -> `Config("aws").get("region")`

Also default values like this:
```typescript
config.getNumber("replicas") || 2,
```
are currently impossible

pawelprazak:
Should be fixed by #287, will re-test

PR #210, opened 2023-10-03
Update docs website

pawelprazak:
- added Basics page
- improved linking between pages
- minor improvements to most of the pages
- missing.md update
- address #211

### DIFF
--- a/website/docs/basics.md
+++ b/website/docs/basics.md
@@ -18,6 +18,11 @@ Pulumi is a registered trademark of [Pulumi Corporation](https://pulumi.com).
 Besom is a Pulumi SDK for Scala 3. It allows you to use Scala to define your infrastructure
 in a type-safe and functional way.
 
+:::tip
+Please pay attention to your dependencies, **only use `org.virtuslab::besom-*`** and not `com.pulumi:*`.<br/>
+Besom **does NOT depend on Pulumi Java SDK**, it is a completely separate implementation.
+:::
+
 ### Concepts
 
 It is important to understand the basic concepts of Pulumi before we dive into the details of Besom.
--- a/website/docs/basics.md
+++ b/website/docs/basics.md
@@ -165,11 +165,21 @@ Output transformations available in Besom:
 - [`map` and `flatMap`](apply_methods) methods take a callback that receives the plain value, and computes a new output
 - [lifting](lifting) directly read properties off an output value
 - [interpolation](interpolator) concatenate string outputs with other strings directly
+- `sequence` method combines multiple outputs into a single output of a list
+- `zip` method combines multiple outputs into a single output of a tuple
+- `traverse` method transforms a map of outputs into a single output of a map
 
 To create an output from a plain value, use the `Output` constructor, e.g.:
 ```scala
 val hello = Output("hello")
 val world = Output.secret("world")
+```
+
+To transform an output value, use the `map` and `flatMap` methods, e.g.:
+```scala
+val hello = Output("hello").map(_.toUpperCase)
+val world = Output.secret("world")
+val helloWorld: Output[String] = hello.flatMap(h => h + "_" + world.map(_.toUpperCase))
 ```
 
 If you have multiple outputs of the same type and need to use them together **as a list** you can use 
@@ -178,7 +188,7 @@ If you have multiple outputs of the same type and need to use them together **as
 ```scala
 val port: Output[String] = pod.name
 val host: Output[String] = node.hostname
-val hello = List(host, port).sequence // we use the extension method here
+val hello: Output[List[String]] = List(host, port).sequence // we use the extension method here
 ```
 
 If you have multiple outputs of different types and need to use them together **as a tuple** you can use the standard 
@@ -187,7 +197,8 @@ If you have multiple outputs of different types and need to use them together **
 ```scala
 val port: Output[Int] = pod.port
 val host: Output[String] = node.hostname
-val hello = host.zip(port).map { case (a, b) => s"https://$hostname:$port/" }
+val hello: Output[(String, Int)] = host.zip(port)
+val url: Output[String] = hello.map { case (a, b) => s"https://$hostname:$port/" }
 ```
 
 If you have a map of outputs and need to use them together **as a map** you can use
--- a/website/docs/basics.md
+++ b/website/docs/basics.md
@@ -135,7 +135,7 @@ To create an output from a plain value, use the `Output` constructor, e.g.:
 val hello = Output("hello")
 ```
 
-If you have multiple outputs and need to use them together you can use the standard 
+If you have multiple outputs and need to use them together as a tuple you can use the standard 
 [`zip`](https://scala-lang.org/api/3.x/scala/collection/View.html#zip-1dd) 
 method to combine them into a single output:
 
@@ -194,7 +194,7 @@ Providers can be configured using [provider configuration](https://www.pulumi.co
 
 :::tip
 It is recommended to [disable default providers](https://www.pulumi.com/blog/disable-default-providers/) 
-if not for [all providers, at least for Kubernetes]((https://www.pulumi.com/docs/concepts/config/#pulumi-configuration-options)).
+if not for [all providers, at least for Kubernetes](https://www.pulumi.com/docs/concepts/config#pulumi-configuration-options).
 :::
 
 #### State
--- a/website/docs/basics.md
+++ b/website/docs/basics.md
@@ -1,0 +1,206 @@
+---
+title: Pulumi Basics
+---
+import Version from '@site/src/components/Version';
+
+Before we dive into the [details of Besom](architecture), let's take a look at the basics of Pulumi.
+This page offers an executive summary of [Pulumi's concepts](https://www.pulumi.com/docs/concepts/).
+
+### What is Pulumi?
+
+Pulumi is a modern infrastructure as code platform. It leverages existing programming languages 
+and their native ecosystem to interact with cloud resources through the Pulumi SDK.
+
+Pulumi is a registered trademark of [Pulumi Corporation](https://pulumi.com).
+
+### What is Besom?
+
+Besom is a Pulumi SDK for Scala 3. It allows you to use Scala to define your infrastructure
+in a type-safe and functional way.
+
+### Concepts
+
+It is important to understand the basic concepts of Pulumi before we dive into the details of Besom.
+We strongly advise to get acquainted with [Pulumi's concepts](https://www.pulumi.com/docs/concepts/)
+documentation as all of that information applies to Besom as well.
+
+Pulumi uses [programs](#programs) to define [resources](#resources) that are managed using [providers](#providers)
+and result in [stacks](#stacks). 
+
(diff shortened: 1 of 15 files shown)


website/docs/basics.md
```diff
@@ -0,0 +1,214 @@
...
+provides the implementation details. If not specified explicitly, the default provider is used.
+Providers can be configured using [provider configuration](https://www.pulumi.com/docs/concepts/resources/providers/#explicit-provider-configuration).
+
+#### Inputs and Outputs
+
+Inputs and Outputs are the primary [asynchronous data types in Pulumi](https://www.pulumi.com/docs/concepts/inputs-outputs/), 
+and they signify values that will be provided by the engine later, when the resource is created and its properties can be fetched.
+`Input[A]` type is an alias for `Output[A]` type used by [resource](#resources) arguments.
+
+Outputs are values of type `Output[A]` and behave very much like [monads](https://en.wikipedia.org/wiki/Monad_(functional_programming)).
```
pawelprazak, line 118: "like" monads or Besom Output is an actual monad?

website/docs/basics.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+If you have multiple outputs and need to use them together you can use the standard 
+[`zip`](https://scala-lang.org/api/3.x/scala/collection/View.html#zip-1dd) 
+method to combine them into a single output:
+
+```scala
+val port: Output[Int] = pod.port
+val host: Output[String] = node.hostname
+val hello = host.zip(port).map((a, b) => s"https://$hostname:$port/")
```
pawelprazak, line 145: not 100% sure about the syntax

website/docs/basics.md
```diff
@@ -0,0 +1,214 @@
...
+- [`pulumi config`](https://www.pulumi.com/docs/concepts/config/#setting-and-getting-configuration-values) command
+
+##### Accessing Configuration from Code
+
+Configuration values can be [accessed](https://www.pulumi.com/docs/concepts/config/#code) from [programs](#programs) 
+using the `Config` object, e.g.:
+
+```scala
+val a = config.get("aws:region")
+val b = Config("aws").map(_.get("region"))
```
pawelprazak, line 179: #204

website/docs/basics.md
```diff
@@ -0,0 +1,214 @@
...
+
+##### Stack References
+
+[Stack References](https://www.pulumi.com/docs/concepts/stack/#stackreferences) allow you to use outputs from other 
+[stacks](#stacks) in your [program](#programs).
+
+To reference values from another stack, create an instance of the `StackReference` type using the fully qualified 
+name of the stack as an input, and then read exported stack outputs by their name.
+
+`StackReference` is not implemented yet in Besom, coming soon.
```
pawelprazak, line 90: #209

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+Resources represent the fundamental units that make up your infrastructure, such as a compute instance,
+a storage bucket, or a Kubernetes cluster.
+
+Resources are defined using a [**resource constructor**](constructors). Each resource in Pulumi has:
+- a [logical name and a physical name](https://www.pulumi.com/docs/concepts/resources/names/#resource-names)
+  The logical name establishes a notion of identity within Pulumi, and the physical name is used as identity by the provider
+- a [resource type](https://www.pulumi.com/docs/concepts/resources/names/#types), which identifies the provider and the kind of resource being created
+- [URN](https://www.pulumi.com/docs/concepts/resources/names/#types), which is an automatically constructed globally unique identifier for the resource.
+
+Each resource also can have:
```
KacperFKorban, line 140: typo: `can also have`

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+- a set of [arguments](https://www.pulumi.com/docs/concepts/resources/properties/) that define the behavior of the resulting infrastructure
+- and a set of [options](https://www.pulumi.com/docs/concepts/options/) that control how the resource is created and managed by the Pulumi engine.
+
+Every is managed by a [provider](#providers) which is a plugin that
```
KacperFKorban, line 144: typo: `resource`?

website/docs/basics.md
KacperFKorban: Hmmm, `Output` has a bit of a complicated semantic. I think that it wouldn't be a crime to call them monads. The only benefit we get from calling it a monad is the analogy. I don't see anyone using generic functions over Monads for `Output`.
Maybe a useful analogy would be to say that it is besom's effect type?

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+val host: Output[String] = node.hostname
+val hello = host.zip(port).map { case (a, b) => s"https://$hostname:$port/" }
+```
+
+If you have a map of outputs and need to use them together **as a map** you can use
+`Output.traverse` method to combine them into a single output:
+
+```scala
+val m: Map[String, Output[String]] = Map(pod.name -> pod.port)
+val o: Output[Map[String, String]] = m.traverse // we use the extension method here
```
KacperFKorban, line 198: minor: Hmmm, I think that it's a very unconventional way of using `treverse`. A `List` example will be more illustrative.

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+Outputs are used to:
+- automatically captures dependencies between [resources](#resources)
+- provide a way to express transformations on its value before it's known
+- deffer the evaluation of its value until it's known
+- track the _secretness_ of its value
+
+Output transformations available in Besom:
+- [`map` and `flatMap`](apply_methods) methods take a callback that receives the plain value, and computes a new output
+- [lifting](lifting) directly read properties off an output value
+- [interpolation](interpolator) concatenate string outputs with other strings directly
```
KacperFKorban, line 167: nit: I would also just mention the methods from below here

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+val world = Output.secret("world")
+```
+
+If you have multiple outputs of the same type and need to use them together **as a list** you can use 
+`Output.sequence` method to combine them into a single output:
+
+```scala
+val port: Output[String] = pod.name
+val host: Output[String] = node.hostname
+val hello = List(host, port).sequence // we use the extension method here
```
KacperFKorban, line 181: nit: It would be super useful to add a type ascription here

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+with [Pulumi Service](https://www.pulumi.com/docs/intro/cloud-providers/pulumi-service/) being the default.
+
+State is used to:
+- track [resources](#resources) that are created by your [program](#programs)
+- record the relationship between resources
+- store metadata about your [project](#projects) and [stacks](#stacks)
+- and store [configuration](#configuration) and secret values
+- and store [stack outputs](#stack-outputs)
+
+:::tip
```
KacperFKorban, line 265: nit: Maybe downgrade this to `note`?

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+Pulumi is a registered trademark of [Pulumi Corporation](https://pulumi.com).
+
+### What is Besom?
+
+Besom is a **Pulumi SDK for Scala 3**. It allows you to use Scala to define your infrastructure
+in a type-safe and functional way.
+
+Besom **does NOT depend on Pulumi Java SDK**, it is a completely separate implementation.
+
+:::tip
```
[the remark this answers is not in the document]
pawelprazak:
```suggestion
:::caution
```

website/docs/basics.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+#### Programs
+
+A Pulumi program, written in a general-purpose programming language, is a collection of [resources](#resources) 
+that are deployed to a [stack](#stacks).
+
+A minimal Besom program consists of:
+* `project.scala` - the program dependencies (here we use [Scala-CLI directives](https://scala-cli.virtuslab.org/docs/guides/using-directives/))
+    ```
```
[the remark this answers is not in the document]
pawelprazak: I've tried that, but it got rendered light gray for some reason and I thought it was a bit too light, but I'm happy to change this.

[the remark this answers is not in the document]
pawelprazak:
```suggestion
    ```scala
```

website/docs/basics.md
pawelprazak:
```suggestion
Each resource can also have:
```

website/docs/basics.md
pawelprazak:
```suggestion
Every [resource](#resources) is managed by a [provider](#providers) which is a plugin that
```

website/docs/basics.md
pawelprazak: OK then, I'll leave as-is, this is the language Pulumi is using in their docs "like monad"

website/docs/basics.md
pawelprazak:
```suggestion
val hello: Output[List[String]] = List(host, port).sequence // we use the extension method here
```

website/docs/basics.md
pawelprazak: I'm sorry, I'm not sure hot to improve this line. I'll leave as-is in the interest of delivering this content to the users ASAP. Please let me know how can I make this line better.

website/docs/basics.md
pawelprazak:
```suggestion
:::note
```

website/docs/basics.md
KacperFKorban: it looks correct
maybe we should consider using mdoc to type-check the snippets?

website/docs/basics.md
KacperFKorban: I would suggest sth like:
```scala
val names: List[String] = List("John", "Paul")
val outputNames: Output[List[String]] = names.traverse(name => Output(name))
```

website/docs/basics.md
pawelprazak: Checking snippets sounds great! Yes, there was a typo.

This compiles:
```
  val port: Output[Int] = Output(8080)
  val host: Output[String] = Output("localhost")
  val hello = host.zip(port).map((hostname, port) => s"https://$hostname:$port/")
```

I've also improved the snippet a bit.

KacperFKorban (approved):

website/docs/basics.md
pawelprazak: Actually, the map example is what users might be looking for (other SDKs tend to support them), but I didn't I can use traverse like that so this is also very useful, thank you

website/docs/basics.md
pawelprazak: done :)

website/docs/basics.md
pawelprazak: I've change the snipped to be more clear:
```
val port: Output[Int] = pod.port
val host: Output[String] = node.hostname
val hello: Output[(String, Int)] = host.zip(port)
val url: Output[String] = hello.map { case (a, b) => s"https://$hostname:$port/" }
```

Issue #198, opened 2023-10-01
Codegen needs to not mangle urn or id fields


lbialy:
Codegen currently creates the full list of fields for a struct before applying mangling to it and so it can yield a `urn_` and `id_` field names. This is incorrect, these fields should join the already mangled list. This is related to the ovh `urn` field fiasco but it's a general logic problem in codegen that needs fixing.

lbialy:
In current codegen it is impossible for mangling to yield `urn_` or `id_` because they are not listed as protected names. This issue should only lead to codegen rejecting schemas where resources define fields called `urn` or `id`.

pawelprazak:
Fixed by #322

PR #217, opened 2023-10-04
Unmangle restricted field names in decoders, fix codecs for enums

### LINKED ISSUE #194: We use `name` instead of `value` when a provider enum is passed
We use `name` instead of `value` when a provider enum is passed, e.g.:
```
  val server = ec2.Instance(
    "web-server-www",
    ec2.InstanceArgs(
      ami = "ami-01342111f883d5e4e",
      instanceType = ec2.enums.InstanceType.T2_Micro, // <---
      vpcSecurityGroupIds = List(securityGroup.id),
      ...
```
resulting in:
```
     Type                      Name                    Status                  Info
 +   pulumi:pulumi:Stack       aws-java-webserver-dev  **creating failed**     1 error
 +   ├─ aws:ec2:SecurityGroup  web-secgrp              created (2s)
 +   └─ aws:ec2:Instance       web-server-www          **creating failed**     1 error


Diagnostics:
  pulumi:pulumi:Stack (aws-java-webserver-dev):
    error: update failed

  aws:ec2:Instance (web-server-www):
    error: 1 error occurred:
    	* creating EC2 Instance: InvalidParameterValue: Invalid value 'T2_Micro' for InstanceType.
    	status code: 400, request id: 3c0be295-7ee8-4649-9e38-7b9a30b4c2ff
```

Workaround is to use:
```
instanceType = ec2.enums.InstanceType.T2_Micro.value
```

### LINKED ISSUE #197: Mangled properties are not resolved by ResourceDecoder
Codegen mangles property names that are in the set of reserved keywords for both JVM, Scala and Pulumi (`wait`, `notify`, `given`, `urn`, `id`, etc). These properties are generated with a following underscore so: `notify_`, `urn_`. We need to implement identical logic (probably via a separate module?) for mangling and demangling between codegen and core and use it in ResourceDecoder field lookup.

KacperFKorban:
closes #197
closes #194

### DIFF
--- a/core/src/main/scala/besom/internal/codecs.scala
+++ b/core/src/main/scala/besom/internal/codecs.scala
@@ -70,6 +70,29 @@ trait Decoder[A]:
       }
     override def mapping(value: Value, label: Label): B = ???
 
+object NameUnmangler:
+  private val mangledAnyRefMethodNames = Set(
+    "eq",
+    "ne",
+    "notify",
+    "notifyAll",
+    "synchronized",
+    "wait",
+    "asInstanceOf",
+    "clone",
+    "equals",
+    "getClass",
+    "hashCode",
+    "isInstanceOf",
+    "toString",
+  ).map(_ + "_")
+
+  /** Keep in sync with `manglePropertyName` in CodeGen.scala */
+  def unmanglePropertyName(name: String): String =
+    if (mangledAnyRefMethodNames.contains(name)) {
+      name.dropRight(1)
+    } else name
+
 object Decoder extends DecoderInstancesLowPrio1:
   import spray.json.*
 
@@ -364,7 +387,7 @@ trait DecoderHelpers:
   inline def derived[A](using m: Mirror.Of[A]): Decoder[A] =
     lazy val labels           = CodecMacros.summonLabels[m.MirroredElemLabels]
     lazy val instances        = CodecMacros.summonDecoders[m.MirroredElemTypes]
-    lazy val nameDecoderPairs = labels.zip(instances)
+    lazy val nameDecoderPairs = labels.map(NameUnmangler.unmanglePropertyName).zip(instances)
 
     inline m match
       case s: Mirror.SumOf[A] => decoderSum(s, nameDecoderPairs)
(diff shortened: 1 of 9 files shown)


core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -70,6 +70,29 @@ trait Decoder[A]:
       }
     override def mapping(value: Value, label: Label): B = ???
 
+object NameUnmangler:
+  private val mangledAnyRefMethodNames = Set(
```
prolativ, line 74: We should also mangle and unmangle property names `urn` and `id`

core/src/test/scala/besom/internal/ValueAssertions.scala
```diff
@@ -0,0 +1,30 @@
...
+trait ValueAssertions:
+  self: munit.Assertions =>
+
+  def assertEqualsMaybeValue(actual: Option[Value], expected: Option[Value]): Unit =
+    (actual, expected) match
+      case (None, None) => ()
+      case (None, Some(Value(Kind.NullValue(_), _))) => ()
+      case (Some(Value(Kind.NullValue(_), _)), None) => ()
+      case (Some(a), Some(b)) => assertEqualsValue(a, b)
+      case _ => throw Exception(s"Values not equal: $actual != $expected")
```
prolativ, line 16: These helpers don't really seem to be helpful IMO. Not only we lose some information about how exactly we (de)serialize things but also in case of an error it's not clear where something was wrong. Relying on `munit`'s `assertEquals` seems a better idea to me as it shows diffs in actual structures of data

core/src/main/scala/besom/types.scala
```diff
@@ -148,30 +146,35 @@ object types:
...
-    given Encoder[E] = Encoder.stringEncoder.contramap(instance => instance.name)
-    given Decoder[E] = Decoder.stringDecoder.emap { (name, label) =>
-      namesToInstances
-        .get(name)
-        .toValidatedOrError(DecodingError(s"$label: `${name}` is not a valid name of `${enumName}`", label = label))
-    }
+    private lazy val valuesToInstances: Map[Any, E] = allInstances.map(instance => instance.value -> instance).toMap
+
+    extension [A](a: A)
+      def asValueAny: Value = a match
```
prolativ, line 166:
```suggestion
      private def asValueAny: Value = a match
```

core/src/main/scala/besom/internal/codecs.scala
prolativ: There's a separate task for that actually https://github.com/VirtusLab/besom/issues/198

PR #230, opened 2023-10-09
Fix GH fork check

pawelprazak:
We seem to be ignoring branch pushes by accident. I hope this will solve the issue, if not we need to dig a little deeper into this.


KacperFKorban (approved):

lbialy:
@pawelprazak this is actually redundant either way because gha does not execute pipelines on PRs from forks, it asks for a permission to run them. CI files are also in CODEOWNERS so can't be changed without explicit approval from the team.

PR #240, opened 2023-10-12
Fix #221 - markdown snippets check with mdoc

### LINKED ISSUE #221: Find a way to typecheck the snippets in markdown
Ideally we'd type check the snippets in readme's and on the website.

pawelprazak:
as in title

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -310,6 +310,20 @@ clean-test-examples:
 	for file in `ls -d examples/*/ | cut -f2 -d'/'`; do just clean-test-example $file; done
 
 ####################
+# Website and docs
+####################
+
+# Runs tests for website and docs
+test-markdown:
+	cs launch org.scalameta:mdoc_2.12:2.3.8 -- --in ./README.md ./CONTRIBUTING.md --out target/mdoc-readme --site.version=$(cat version.txt)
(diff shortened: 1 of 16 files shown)


Justfile
```diff
@@ -309,6 +309,20 @@ test-examples:
...
 	for file in `ls -d examples/*/ | cut -f2 -d'/'`; do just clean-test-example $file; done
 
+####################
+# Website and docs
+####################
+
+# Runs tests for website and docs
+test-markdown:
+	cs launch org.scalameta:mdoc_2.12:2.3.8 -- --in ./README.md ./CONTRIBUTING.md --out target/mdoc-readme --site.version=$(cat version.txt)
+	cs launch org.scalameta:mdoc_2.12:2.3.8 -- --in ./website --out target/mdoc-website --exclude node_modules --site.version=$(cat version.txt)
```
KacperFKorban, line 319: Does it actually test the snippets? Don't we actually need to add the dependencies etc.?
Also shouldn't the snippets use the `mdoc` annotations? I'm not an mdoc expert, but the snippet in the guide has it.

pawelprazak: not an expert on mdco either, but this works :)

KacperFKorban: But did u manage to make it give a compilation error for any snippet?

pawelprazak: yes, it gave an error :)

KacperFKorban (approved):

PR #253, opened 2023-10-18
Improve core integration tests

pawelprazak:
- base integration framework
- refactored core integration test
- updated test deps
- remove fetch-depth: 0 in workflows

Relates to: #237

Should be merged before #234 and #242

PR #256, opened 2023-10-18
Run random-example test programatically

pawelprazak:
On top of #253


pawelprazak:
Thank you for the review :)

PR #262, opened 2023-10-25
Codegen update 3

pawelprazak:
- [Make codegen integration tests cleaner](https://github.com/VirtusLab/besom/pull/262/commits/633815863316c84b044b374c136fe0bf1211ec85)
- [Add test schemas for docker and kubernetes](https://github.com/VirtusLab/besom/pull/262/commits/2c26608c009b73f08949798ee410f9edceaf7ab7)


integration-tests/CodegenTests.test.scala
```diff
@@ -1,8 +1,8 @@
 package besom.integration.codegen
 
-import os./
 import besom.integration.common.*
 import munit.{Slow, TestOptions}
+import os.{/, *}
```
prolativ, line 5: `os.*` imports `os./` anyway. Not sure if we need `*` though as typically the `os.`prefix is always used

prolativ (approved):

integration-tests/CodegenTests.test.scala
pawelprazak:
```suggestion
import os.*
```

PR #265, opened 2023-10-25
Codegen update 6

pawelprazak:
Codegen changes:
- index module will be added when not provided by the schema (but also always removed from package name)
- dash-es will be removed from package and file path
- module parts will be split on `/` in packages and file path (there are schemas that do not provide correct module regexp)

### DIFF
--- a/codegen/src/TypeMapper.test.scala
+++ b/codegen/src/TypeMapper.test.scala
@@ -90,7 +90,7 @@ class TypeMapperTest extends munit.FunSuite {
       language = Language(
         java = Java(
           packages = Map(
-            "authentication.k8s.io/v1" -> "authentication/v1"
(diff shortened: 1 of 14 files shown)


prolativ (approved):

codegen/src/TypeMapper.test.scala
```diff
@@ -114,8 +110,7 @@ class TypeMapperTest extends munit.FunSuite {
             "rbac.authorization.k8s.io/v1" -> "rbac/authorization/v1"
```
prolativ, line 110:
```suggestion
            "rbac.authorization.k8s.io/v1" -> "rbac.authorization.v1"
```

pawelprazak:
Thank you! :)

PR #266, opened 2023-10-26
Update PulumiPackage and warn about missing functionality

pawelprazak:
- update our metaschema model according to upstream JSON Schema (codegen/resources/pulumi.js)
- warn about parts of schema we ignore at the moment
- add kubernetes schema to codegen tests

### DIFF
--- a/codegen/src/Utils.scala
+++ b/codegen/src/Utils.scala
@@ -25,18 +25,24 @@ object Utils {
       pkg => pkg.split("/").filter(_.nonEmpty).toSeq
 
     private def languageModuleToPackageParts: String => Seq[String] = {
-      pulumiPackage.language.java.packages.view
-        .mapValues { pkg =>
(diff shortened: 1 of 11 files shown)


codegen/src/Utils.scala
```diff
@@ -25,18 +25,24 @@ object Utils {
...
-        .mapValues { pkg =>
+      if (pulumiPackage.language.java.packages.view.nonEmpty) {
+        pulumiPackage.language.java.packages.view.mapValues { pkg =>
           pkg.split("\\.").filter(_.nonEmpty).toSeq
-        }
-        .toMap
-        .withDefault(slashModuleToPackageParts)
+        }.toMap
+      } else {
+        // use nodejs mapping a a fallback
```
prolativ, line 33:
```suggestion
        // use nodejs mapping as a fallback
```

prolativ (approved):

PR #268, opened 2023-10-26
Codegen update 9

pawelprazak:
[Fix codegen for 'secrets' test schema](https://github.com/VirtusLab/besom/commit/1398b164c23ee4d194ecb27ec486181008dce60a)

### DIFF
--- a/codegen/src/CodeGen.scala
+++ b/codegen/src/CodeGen.scala
@@ -105,7 +105,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
     sourceFilesForResource(
       typeCoordinates = typeCoordinates,
       resourceDefinition = pulumiPackage.provider,
-      typeToken = s"pulumi:provider:${pulumiPackage.name}",
+      typeToken = TypeToken("pulumi", "provider", pulumiPackage.name),
       isProvider = true
     )
   }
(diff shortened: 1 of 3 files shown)


codegen/src/CodeGen.scala
```diff
@@ -466,7 +466,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
 
     // the type has to match Pulumi's resource type schema, e.g. kubernetes:core/v1:Pod
     // please make sure, it contains 'index' instead of empty module part if needed
-    val typ = Lit.String(typeToken)
+    val typ = Lit.String(typeToken.sanitized.asString)
```
prolativ, line 469: Are we sure it's OK to modify the original unparsed token in any way?

codegen/src/CodeGen.scala
```diff
@@ -459,7 +466,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
...
```
prolativ, line 469: Are we sure it's OK to modify the original unparsed token in any way?

pawelprazak: It is necessary, according to our own SDK validation, I'm not 100% sure the engine requires it. CC @lbialy

pawelprazak:
Rebased on develop

codegen/src/CodeGen.scala
pawelprazak: I have an idea how to correct it. Coming soon.

pawelprazak:
closed in favor of https://github.com/VirtusLab/besom/pull/270

PR #270, opened 2023-10-27
Codegen - deal with broken schemas

pawelprazak:
- [Codegen disallows empty module in Pulumi type](https://github.com/VirtusLab/besom/commit/5b13046fb78e56d726c6f34e574019190d31da64)
- [Add ensureIndexModule to codegen config for optional sanitization](https://github.com/VirtusLab/besom/commit/334208db0e1fd271e64daf62ba6a4c35456eeaa2)
- [Add TypeToken in codegen](https://github.com/VirtusLab/besom/commit/04240276bc816e03c019b239af006bafb1272aee)

other small fixes

### DIFF
--- a/codegen/src/CodeGen.scala
+++ b/codegen/src/CodeGen.scala
@@ -105,7 +105,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
     sourceFilesForResource(
       typeCoordinates = typeCoordinates,
       resourceDefinition = pulumiPackage.provider,
-      typeToken = s"pulumi:provider:${pulumiPackage.name}",
+      typeToken = TypeToken("pulumi", "provider", pulumiPackage.name),
       isProvider = true
     )
   }
@@ -328,7 +328,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
           sourceFilesForResource(
             typeCoordinates = typeMapper.parseTypeToken(typeToken, moduleToPackageParts, providerToPackageParts),
             resourceDefinition = resourceDefinition,
-            typeToken = typeToken,
+            typeToken = TypeToken(typeToken),
             isProvider = false
           )
       }
@@ -339,7 +339,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
   def sourceFilesForResource(
     typeCoordinates: PulumiTypeCoordinates,
     resourceDefinition: ResourceDefinition,
-    typeToken: String,
+    typeToken: TypeToken,
     isProvider: Boolean
   ): Seq[SourceFile] = {
     val baseClassCoordinates = typeCoordinates.asResourceClass(asArgsType = false)
@@ -466,7 +466,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
 
     // the type has to match Pulumi's resource type schema, e.g. kubernetes:core/v1:Pod
     // please make sure, it contains 'index' instead of empty module part if needed
-    val typ = Lit.String(typeToken)
+    val typ = Lit.String(typeToken.asString)
 
     val baseCompanion =
       if (hasOutputExtensions) {
--- a/codegen/src/TypeMapper.scala
+++ b/codegen/src/TypeMapper.scala
@@ -36,8 +36,6 @@ object TypeToken {
       `type` = `type`
     )
   }
-
-  def unapply(t: TypeToken): Option[(String, String, String)] = Some((t.provider, t.module, t.`type`))
 }
 
 class TypeMapper(
--- a/codegen/src/TypeMapper.scala
+++ b/codegen/src/TypeMapper.scala
@@ -5,9 +5,7 @@ import scala.meta._
 import scala.meta.dialects.Scala33
 import besom.codegen.metaschema._
 
-import scala.annotation.unused
-
-case class TypeToken private (provider: String, module: String, `type`: String, @unused dummy: Boolean = false) {
+case class TypeToken private (provider: String, module: String, `type`: String) {
   def asString: String = s"${provider}:${module}:${`type`}"
 }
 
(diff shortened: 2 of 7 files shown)


codegen/src/TypeMapper.scala
```diff
@@ -1,35 +1,56 @@
 package besom.codegen
 
+import besom.codegen.Config.ProviderConfig
+
 import scala.util.matching.Regex
 import scala.meta._
 import scala.meta.dialects.Scala33
 import besom.codegen.metaschema._
 
+case class TypeToken private (provider: String, module: String, `type`: String, raw: String) {
```
prolativ, line 10: The design of this class looks slightly suspicious and against the usual conventions about case classes.
I would recommend not to override `toString` - if we use it just for debugging it would be better not to lose the information about the original string. Otherwise one might get surprised that `foo:bar:Baz` and `foo:bar:Baz` are not equal (because the `raw` part was different). If we need to make a string from that as a part of business logic - let's make this explicit by choosing a different method name for clarity.
Also I'm not sure if it's a good idea to use the extractor if all the fields are strings and we rely just on the order of fields - something's likely to go wrong if some code around this class gets refactored - it would be better to rely on named field access

codegen/src/TypeMapper.scala
```diff
@@ -1,35 +1,56 @@
...
+}
+
+object TypeToken {
+  private val typeTokenFmt: String    = "(.*):(.*)?:(.*)" // provider:module:type
+  private val typeTokenPattern: Regex = ("^" + typeTokenFmt + "$").r
+
+  def apply(typeToken: String)(implicit config: ProviderConfig): TypeToken = typeToken match {
+    case typeTokenPattern(p, m, t) => {
+      val module =  m match {
+        case "" if config.ensureIndexModule => "index"
```
prolativ, line 22: Shouldn't it be `enforceIndexModule`? `ensure` would rather suggest crashing if the requirement is not fulfilled. Or maybe `inferIndexModule` would be even better?

codegen/src/CodeGen.scala
```diff
@@ -466,7 +466,7 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
 
     // the type has to match Pulumi's resource type schema, e.g. kubernetes:core/v1:Pod
     // please make sure, it contains 'index' instead of empty module part if needed
-    val typ = Lit.String(typeToken)
+    val typ = Lit.String(typeToken.asString)
```
prolativ, line 469: Shouldn't we use `typeToken.raw` here?

codegen/src/TypeMapper.scala
pawelprazak: ensure - make certain that (something) will occur or be the case

there is no inference, just making sure the index is present

codegen/src/CodeGen.scala
pawelprazak: yes, this is better, thank you

codegen/src/TypeMapper.scala
pawelprazak: I'll make a change to `enforceIndexModule` to avoid iny confusion, thank you

codegen/src/TypeMapper.scala
pawelprazak: Thank you I will remove the `toString` and make the `apply` always create a correct object

codegen/src/CodeGen.scala
pawelprazak: on a second thought, if we use `raw` we cannot add the index conditionally, because the module will be empty anyway

prolativ: I think the point is we shouldn't add `index` here by ourselves in any case

codegen/src/TypeMapper.scala
```diff
@@ -1,35 +1,69 @@
...
 import besom.codegen.metaschema._
 
+import scala.annotation.unused
+
+case class TypeToken private (provider: String, module: String, `type`: String, @unused dummy: Boolean = false) {
+  def asString: String = s"${provider}:${module}:${`type`}"
+}
+
+object TypeToken {
+  private val typeTokenFmt: String    = "(.*):(.*)?:(.*)" // provider:module:type
```
prolativ, line 17: `?` seems redundant here as `.*` would accept an empty string anyway

pawelprazak: yes, you're probably right, but this one better states the intention IMO

codegen/src/TypeMapper.scala
```diff
@@ -5,6 +5,43 @@ import scala.meta._
 import scala.meta.dialects.Scala33
 import besom.codegen.metaschema._
 
+import scala.annotation.unused
+
+case class TypeToken private (provider: String, module: String, `type`: String, @unused dummy: Boolean = false) {
```
prolativ, line 10:
```suggestion
case class TypeToken (provider: String, module: String, `type`: String) {
```
and remove the 3-parameter overload of `apply`

codegen/src/TypeMapper.scala
```diff
@@ -5,6 +5,43 @@ import scala.meta._
...
+
+  def apply(provider: String, module: String, `type`: String): TypeToken = {
+    new TypeToken(
+      provider = provider,
+      module = enforceNonEmptyModule(module),
+      `type` = `type`
+    )
+  }
+
+  def unapply(t: TypeToken): Option[(String, String, String)] = Some((t.provider, t.module, t.`type`))
```
prolativ, line 42: This should be redundant when the public main constructor gets changed to have 3 parameters

codegen/src/TypeMapper.scala
pawelprazak: the overridden apply has `enforceNonEmptyModule(module),`, so I'm afraid it might be necessary

pawelprazak: but thank you for noticing the dummy is not necessary after all :)

codegen/src/TypeMapper.scala
pawelprazak: thank you! :)

prolativ (approved):

PR #274, opened 2023-10-31
Generalize type token

prolativ:
Pulumi tokens can refer not only to types but also to functions


pawelprazak (approved):
I really like the new names, they are immediately recognizable for anyone who knows Pulumi internals :)

I appreciate the code cleanup as well.

codegen/src/TypeMapper.scala
```diff
@@ -95,21 +48,25 @@ class TypeMapper(
     val uniformedTypeToken = typeToken.toLowerCase
 
     val typeCoordinates =
-      parseTypeToken(typeToken, packageInfo.moduleToPackageParts, packageInfo.providerToPackageParts)
+      PulumiDefinitionCoordinates.fromRawToken(
```
pawelprazak, line 51: I like this very much, a factory method instead of an util :)

PR #275, opened 2023-10-31
No imports in generated code

prolativ:
Avoid name clashes with custom classes from schemas

* Don't use imports in generated code
* Extract some codegen helper logic


codegen/src/CodeGen.scala
```diff
@@ -16,11 +16,6 @@ class CodeGen(implicit providerConfig: Config.ProviderConfig, typeMapper: TypeMa
 
   import CodeGen._
 
-  val commonImportedIdentifiers = Seq(
```
pawelprazak, line 19: good, we were clashing with provider code with prod schemas

pawelprazak (approved):
LGTM, thank you!

PR #280, opened 2023-11-06
Codegen all packages

pawelprazak:
Use Codegen schema download as universal mechanism across all parts that require schemas and codegen (provider generators, integration tests, etc.)


codegen/src/PackageMetadata.scala
```diff
@@ -0,0 +1,75 @@
...
+  def apply(): PackageVersion = new PackageVersion(DefaultVersion)
+
+  def isValid(version: String): Boolean = SemverRegex.matches(version)
+
+  implicit class PackageVersionOps(version: PackageVersion)(implicit logger: Logger) {
+    def isDefault: Boolean = version == DefaultVersion
+
+    def reconcile(otherVersion: PackageVersion): PackageVersion = {
+      (version, otherVersion) match {
+        case (version, _) if !version.isDefault => version
```
prolativ, line 66: What is the expectation if both versions are nondefault? Should the former one be taken and the latter discarded without a warning? The intent seems quite unclear to me

codegen/src/Main.scala
```diff
@@ -1,95 +1,122 @@
...
+          schemaOrMetadata = Right(PackageMetadata(name, version))
+        )
+      case "metadata" :: metadataPath :: Nil =>
+        generator.generatePackageSources(Right(PackageMetadata.fromJsonFile(os.Path(metadataPath))))
+      case "metadata" :: metadataPath :: outputDir :: Nil =>
+        implicit val codegenConfig: CodegenConfig = CodegenConfig(outputDir = Some(os.rel / outputDir))
+        generator.generatePackageSources(Right(PackageMetadata.fromJsonFile(os.Path(metadataPath))))
+      case "schema" :: providerName :: schemaVersion :: Nil =>
+        generator.generatePackageSources(
+          schemaOrMetadata = Right(PackageMetadata(providerName, schemaVersion))
```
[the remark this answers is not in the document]
pawelprazak: it can be either a path to schema file or metadata (name, ver, server)

codegen/src/Main.scala
```diff
@@ -1,95 +1,122 @@
...
-    besomVersion: String,
-    preLoadSchemas: Map[(ProviderName, SchemaVersion), os.Path] = Map()
-  ): os.Path = {
-    implicit val providerConfig: Config.ProviderConfig = Config.providersConfigs(providerName)
-    implicit val logger: Logger                        = new Logger
-
-    val schemaProvider = new DownloadingSchemaProvider(schemaCacheDirPath = schemasDir)
-    val outputDir      = codegenDir / providerName / schemaVersion
+    schemaOrMetadata: Either[SchemaFile, PackageMetadata]
+  )(implicit config: CodegenConfig = CodegenConfig()): Result = {
```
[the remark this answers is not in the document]
pawelprazak: thank you, I was not aware of the problems it might cause, I will correct :)

codegen/src/Config.scala
```diff
@@ -1,6 +1,24 @@
 package besom.codegen
 
+// noinspection ScalaWeakerAccess
 object Config {
+
+  val DefaultJavaVersion = "11"
+  val DefaultScalaVersion = "3.3.1"
+
+  val DefaultBesomVersion: String = os.read(os.pwd / "version.txt").trim
```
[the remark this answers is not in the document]
pawelprazak: you are right, esp. in context ot using codegen as an executable jar, I will correct this

codegen/src/PackageMetadata.scala
pawelprazak: would changing the name of the parameter from `otherVersion` to `defaultVersion` would make the intent more clear, or renaming the method to `versionOrDefault`?

the intent is handling missing version in schema file

codegen/src/Main.scala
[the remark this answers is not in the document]
prolativ: Sorry, I missed `ever` in the sentence above 😅 I was simply confused that for `"schema"` it's not 
```scala
generatePackageSources(Left(...))
```
with a schema but instead `Right(PackageMetadata(...))` is passed explicitly here

codegen/src/PackageMetadata.scala
prolativ: I would rather expect something like
```scala
      (version.isDefault, otherVersion.isDefault) match {
        case (true, false) => version
        case (false, true) => otherVersion
        case (true, true) => PackageVersion(DefaultVersion)
        case (false, false) => ??? // Fixme: which version to choose?
```

pawelprazak:
```
    def reconcile(otherVersion: PackageVersion): PackageVersion = {
      (version.isDefault, otherVersion.isDefault) match {
        case (true, false) => version
        case (false, true) => otherVersion
        case (false, false) => version // we ignore the other version
        case (true, true)  =>
          // this code should be only reached in tests
          logger.warn(s"Using default version '${DefaultVersion}' for the package - suitable only for tests")
          PackageVersion()
      }
    }
    ```

File: codegen/src/Config.scala (hunk 4 above)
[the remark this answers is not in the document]
Comment (pawelprazak) [reply codegen/src/Config.scala:9]: What if we probe if the file exists, and throw an actionable error message? 
I'd like to preserve the convention-over-configuration properties without shooting ourselves in the proverbial foot ;)

File: codegen/src/PackageMetadata.scala [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+  def apply(): PackageVersion = new PackageVersion(DefaultVersion)
+
+  def isValid(version: String): Boolean = SemverRegex.matches(version)
+
+  implicit class PackageVersionOps(version: PackageVersion)(implicit logger: Logger) {
+    def isDefault: Boolean = version == DefaultVersion
+
+    def reconcile(otherVersion: PackageVersion): PackageVersion = {
+      (version.isDefault, otherVersion.isDefault) match {
+        case (true, false) => version
```
prolativ, line 66: My bad, I inverted the condition - we should choose the NON-default version (with `isDefault` set to `false`) 😅

pawelprazak: right! 😅

codegen/src/SchemaProvider.scala
```diff
@@ -2,89 +2,144 @@ package besom.codegen
...
-      val schemaSource = s"$providerName@$schemaVersion"
-      os.makeDir.all(schemaFilePath / os.up)
-      os.proc("pulumi", "package", "get-schema", schemaSource).call(stdout = schemaFilePath)
+      logger.debug(
+        s"Downloading schema for '${metadata.name}:${metadata.version}' into '${schemaFilePath.relativeTo(os.pwd)}'"
+      )
+
+      val schemaSource =
+        if (metadata.version.isDefault)
+          metadata.name
```
prolativ, line 34: I have already experienced some problems caused by not specifying the exact version for a schema so I think we should avoid that

codegen/src/SchemaProvider.scala
```diff
@@ -2,89 +2,144 @@ package besom.codegen
...
+    // try to get fallback version information from the schema file path
+    val pathVersion = {
+      val directory = (schema / os.up).last
+      if (PackageVersion.isValid(directory)) PackageVersion(directory) else PackageVersion()
+    }
+
+    // parse and save the schema using path corrected for the actual name and version for the package
+    val pulumiPackage         = PulumiPackage.fromFile(schema)
+    val metadata              = pulumiPackage.toPackageMetadata(pathVersion)
+    val correctSchemaFilePath = schemaCacheDirPath / metadata.name / metadata.version / schemaFileName
```
prolativ, line 78: In what sense is this path `correct`? Is there any other path that is incorrect?

pawelprazak: there can be (and is often in tests) a defference of name and version between the arguments that are provided and the contents of the actual schema file e.g. multpile test schemas from upstream have name 'example' and no version

codegen/src/SchemaProvider.scala
pawelprazak: what kind of problems? some schemas does not have version inside

codegen/src/SchemaProvider.scala
```diff
@@ -2,89 +2,144 @@ package besom.codegen
...
+      )
+
+      val schemaSource =
+        if (metadata.version.isDefault)
+          metadata.name
+        else
+          s"${metadata.name}@${metadata.version}"
+
+      val installCmd: List[String] =
+        List("pulumi", "plugin", "install", "resource", metadata.name, metadata.version) ++
```
pawelprazak, line 39: I've just realize that this will not work correctly for "0.0.0" aka default version

pawelprazak: Done

codegen/src/SchemaProvider.scala
pawelprazak: I'd rather try to resolve the latest version, I've corrected the logic to behave that way

codegen/src/PackageMetadata.scala
pawelprazak: done :)

codegen/src/SchemaProvider.scala
```diff
@@ -2,89 +2,149 @@ package besom.codegen
...
+      val schemaSource =
+        if (metadata.version.isDefault)
+          metadata.name
+        else
+          s"${metadata.name}@${metadata.version}"
+
+      val installCmd: List[String] =
+        List("pulumi", "plugin", "install", "resource", metadata.name) ++ {
+          // use version only if it is not the default, otherwise try to install the latest
+          if (metadata.version.isDefault) List(metadata.version) else List.empty
```
prolativ, line 41:
```suggestion
          if (!metadata.version.isDefault) List(metadata.version) else List.empty
```
Otherwise the implementation seems to be against the comment.
But if the default version is `0.0.0`, which is only used in codegen tests, without the intent to actually download this schema, maybe `installCmd` shouldn't be run for the default version at all?

prolativ (approved):

PR #281, opened 2023-11-07
Codegen - fix tests

pawelprazak:
Fixes on top of #280

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -28,6 +28,8 @@ scala-cli-test-options-zio := scala-cli-main-options-zio + " " + scala-cli-cover
 
 publish-maven-auth-options := "--user env:OSSRH_USERNAME --password env:OSSRH_PASSWORD --gpg-key $PGP_KEY_ID --gpg-option --pinentry-mode --gpg-option loopback --gpg-option --passphrase --gpg-option $PGP_PASSWORD"
 
+scala-cli-bloop-opts := "--jvm=17 --bloop-jvm=17 --bloop-java-opt='-XX:MaxRAMPercentage=95.0' --bloop-java-opt='-XX:InitialRAMPercentage=70.0' --bloop-java-opt='-XX:+UseUseParallelGC'"
+
 # This list of available targets
 default:
 	@just --list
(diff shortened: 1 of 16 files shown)


pawelprazak:
rebased on dependent PR

pawelprazak:
I've changed the PR base for easier review

codegen/src/Utils.scala
```diff
@@ -33,13 +39,13 @@ object Utils {
       if (pulumiPackage.language.java.packages.view.nonEmpty) {
         pulumiPackage.language.java.packages.view.mapValues { pkg =>
           pkg.split("\\.").filter(_.nonEmpty).toSeq
-        }.toMap
+        }.toMap.withDefault(slashModuleToPackageParts) // fallback to slash mapping
```
pawelprazak, line 42: I don't know why, but before this change, we got key not found error instead of the fallback function

Justfile
```diff
@@ -257,11 +262,11 @@ generate-provider-sdk schema-name schema-version:
 
 # Compiles the previously generated scala API code for the given provider, e.g. `just compile-provider-sdk kubernetes 4.0.0`
 compile-provider-sdk schema-name schema-version:
-	scala-cli --power compile {{codegen-output-dir}}/{{schema-name}}/{{schema-version}} --suppress-experimental-feature-warning
+	scala-cli --power compile {{scala-cli-bloop-opts}} {{codegen-output-dir}}/{{schema-name}}/{{schema-version}} --suppress-experimental-feature-warning --interactive=false
```
lbialy, line 265: these opts will be ignored if bloop is already running

codegen/src/Utils.scala
lbialy: why wouldn't we?

lbialy (approved):
just one nitpick in comments

codegen/src/Utils.scala
pawelprazak: IDK, I just fond this one confusing

Justfile
pawelprazak: should we `scala-cli bloop exit` before compiling then, WDYT?

pawelprazak:
thank you for the review! regarding bloop exit, we need more work on the mass-generation of the package anyway, so in a followup I'll try to further optimize this problem

Issue #208, opened 2023-10-02
Add strongly typed configuration


pawelprazak:
We are missing [Strongly Typed](https://www.pulumi.com/docs/concepts/config/#structured-configuration) configuration

Relates to #204

pawelprazak:
After closer inspection of the upstream docs, this feature looks CLI specific and does not require support from language host. Closing.

Issue #207, opened 2023-10-02
Add structured configuration


pawelprazak:
We are missing [Structured Configuration](https://www.pulumi.com/docs/concepts/config/#structured-configuration) in the API

Add:
- `def requireJson(key: NonEmptyString): Output[JsValue]`
- `def requireObject[A: JsonFormat](key: (non-empty?)String): Output[A] = requireJson(key).map(_.convertTo[A])`

Relates to #204

PR #287, opened 2023-11-10
Fix config in SDK

prolativ:


core/src/main/scala/besom/internal/Config.scala
```diff
@@ -168,6 +124,25 @@ object Config:
...
+    def requireString(key: NonEmptyString)(using Context): Output[String] = output.flatMap(_.requireString(key))
+
+    def getDouble(key: NonEmptyString)(using Context): Output[Option[Double]] = output.flatMap(_.getDouble(key))
+    def requireDouble(key: NonEmptyString)(using Context): Output[Double] = output.flatMap(_.requireDouble(key))
+    
+    def getInt(key: NonEmptyString)(using Context): Output[Option[Int]] = output.flatMap(_.getInt(key))
+    def requireInt(key: NonEmptyString)(using Context): Output[Int] = output.flatMap(_.requireInt(key))
+
+    def getBoolean(key: NonEmptyString)(using Context): Output[Option[Boolean]] = output.flatMap(_.getBoolean(key))
+    def requireBoolean(key: NonEmptyString)(using Context): Output[Boolean] = output.flatMap(_.requireBoolean(key))
```
prolativ, line 143: Should we also supports maps and lists now?

pawelprazak: I think [this might be what you are referring to](https://www.pulumi.com/docs/concepts/config/#structured-configuration), and yes, we'd like to support this in the future.

core/src/main/scala/besom/util/NonEmptyString.scala
```diff
@@ -3,16 +3,17 @@ package besom.util
...
-
 opaque type NonEmptyString <: String = String
 
 object NonEmptyString:
   def apply(s: String): Option[NonEmptyString] =
     if s.isBlank then None else Some(s)
 
+  given nonEmptyStringOps: {} with
+    extension (nes: NonEmptyString)
+      inline def +++(other: String): NonEmptyString = nes + other
```
pawelprazak, line 14: can we name it? 3 plus signs are hard understand IMO

core/src/main/scala/besom/internal/Config.scala
```diff
@@ -6,117 +6,64 @@ import scala.util.Try
...
   val namespace: NonEmptyString,
-  private val isProjectName: Boolean,
-  private val configMap: Map[String, String],
-  private val configSecretKeys: Set[String]
+  private[internal] val isProjectName: Boolean,
+  private[internal] val configMap: Map[NonEmptyString, String],
+  private[internal] val configSecretKeys: Set[NonEmptyString]
 ):
-  private def fullKey(key: String): String = namespace + ":" + key
+  private def fullKey(key: NonEmptyString): NonEmptyString = namespace +++ ":" +++ key
```
pawelprazak, line 13: wouldn't it be more readable to just use string interpolation instead of `namespace +++ ":" +++ key`?

pawelprazak (approved):
had some questions, looks good otherwise

core/src/main/scala/besom/internal/Config.scala
prolativ: That would probably require implementing an interpolator for nonempty strings

core/src/main/scala/besom/internal/Config.scala
prolativ: Structured configuration could also be available to besom users as a case class. That would require to make sure that `ConfigValueReader` is available to users so that they can implement an instance for a desired class. In a perfect case this should be reducible to adding `derives ConfigValueReader` (or whatever we decide to call this trait finally)

core/src/main/scala/besom/util/NonEmptyString.scala
prolativ: `concat`? The point is that this should be just like `++` for normal strings but because of subtyping it doesn't seem possible to reuse `++`

lbialy:
I've made an issue to extend interpolator with NonEmptyStrings.

core/src/main/scala/besom/internal/Config.scala
pawelprazak: this is probably [strongly typed config](https://www.pulumi.com/docs/concepts/config/#strongly-typed-configuration)

we have both on our roadmap: #207 #208

PR #294, opened 2023-11-13
Test and document Config

### LINKED ISSUE #204: Improve `Config` ergonomy

### LINKED ISSUE #205: `config.getSecret` shorthand config key name is broken
The shorthand config key name is broken for secrets, this returns empty output:
`val mySecret = config.getSecret("mySecret")`

and when the key has the explicit namespace it works as expected:
`val mySecret = config.getSecret("aws-secrets-manager:mySecret")`

Relates to #204

pawelprazak:
- finishing touches on #287 to improve `Config` ergonomy
- add `Output[Option[A]].orElse` and `.getOrElse`
- add unit tests, integration tests, scaladoc

Fixes #204
Fixes #205

### DIFF
--- a/core/src/main/scala/besom/future.scala
+++ b/core/src/main/scala/besom/future.scala
@@ -3,13 +3,40 @@ package besom
 import besom.internal.*
 import scala.concurrent.*
 
+/** A module that provides a runtime for the [[scala.concurrent.Future]] monad.
+  */
 trait FutureMonadModule extends BesomModule:
   override final type Eff[+A] = scala.concurrent.Future[A]
-  given ExecutionContext      = scala.concurrent.ExecutionContext.global
+  given ExecutionContext                 = scala.concurrent.ExecutionContext.global
   protected lazy val rt: Runtime[Future] = FutureRuntime()
 
   implicit val toFutureFuture: Result.ToFuture[Eff] = new Result.ToFuture[Future]:
     def eval[A](fa: => Future[A]): () => Future[A] = () => fa
 
+/** Besom API entry point for the [[scala.concurrent.Future]] monad.
+  *
+  * All Pulumi programs are executed in Besom context [[besom.Context]]
+  *
+  * Most notable methods exposed by [[besom.Pulumi]] are:
+  *   - [[besom.internal.BesomModule.run]] - the Pulumi program function
+  *   - [[besom.internal.BesomSyntax.exports]] - the Pulumi Stack outputs
+  *   - [[besom.internal.BesomSyntax.config]] - configuration and secrets
+  *   - [[besom.internal.BesomSyntax.log]] - all your logging needs
+  *
+  * Inside `Pulumi.run` block you can use all methods without `Pulumi.` prefix. All functions that belong to Besom
+  * program but are defined outside the `Pulumi.run` block should have the following using clause: `(using Context)` or
+  * `(using besom.Context)` using a fully qualified name of the type.
+  *
+  * The hello world example:
+  * {{{
+  * import besom.*
+  *
+  * @main def main = Pulumi.run {
+  *   for
+  *     _ <- log.warn("Nothing's here yet, it's waiting for you to write some code!")
+  *   yield exports()
+  * }
+  * }}}
+  */
 object Pulumi extends FutureMonadModule
-export Pulumi.{ *, given }
+export Pulumi.{*, given}
--- a/core/src/main/scala/besom/internal/BesomModule.scala
+++ b/core/src/main/scala/besom/internal/BesomModule.scala
@@ -3,11 +3,44 @@ package besom.internal
 import com.google.protobuf.struct.*
 import besom.internal.logging.{LocalBesomLogger => logger, BesomLogger}
 
+/** An abstract effect Besom module, which can be implemented for different effect types.
+  * @tparam Eff
+  *   the effect type
+  * @see
+  *   [[besom.Pulumi]]
+  * @see
+  *   [[besom.internal.BesomModule]]
+  * @see
+  *   [[besom.internal.BesomSyntax]]
+  */
 trait EffectBesomModule extends BesomSyntax:
   type Eff[+A]
 
   protected lazy val rt: Runtime[Eff]
 
+  /** Run a [[besom.Pulumi]] program in the Besom [[besom.Context]] and export Stack outputs.
+    *
+    * Inside `Pulumi.run` block you can use all methods without `Pulumi.` prefix. All functions that belong to Besom program but are defined
+    * outside the `Pulumi.run` block should have the following using clause: `(using Context)` or `(using besom.Context)` using a fully
+    * qualified name of the type.
+    *
+    * Example:
+    * {{{
+    * import besom.*
+    * import besom.api.aws
+    *
+    * @main def run = Pulumi.run {
+    *   for
+    *     bucket <- aws.s3.Bucket("my-bucket")
+    *   yield exports(
+    *     bucketUrl = bucket.websiteEndpoint
+    *   )
+    * }
+    * }}}
+    *
+    * @param program
+    *   the program to run
+    */
   def run(program: Context ?=> Output[Exports]): Unit =
     val everything: Result[Unit] = Result.scoped {
       for
@@ -36,6 +69,12 @@ trait EffectBesomModule extends BesomSyntax:
       case Right(_) =>
         sys.exit(0)
 
+/** The Besom module provides the core functionality of the Besom runtime.
+  * @see
+  *   [[besom.Pulumi]]
+  * @see
+  *   [[besom.internal.EffectBesomModule]]
+  */
 trait BesomModule extends EffectBesomModule {
   export besom.types.{*, given}
   export besom.util.interpolator.{*, given}
--- a/core/src/main/scala/besom/internal/Output.scala
+++ b/core/src/main/scala/besom/internal/Output.scala
@@ -8,9 +8,9 @@ import scala.collection.BuildFrom
   *
   * Invariant: dataResult has to be registered in TaskTracker by the time it reaches the constructor here!
   * @param dataResult
-  *   Effect of type Result[A]
+  *   Effect of type [[Result]][A]
   * @param ctx
-  *   Context
+  *   the Besom [[Context]]
   */
 class Output[+A] private[internal] (using private[besom] val ctx: Context)(
   private val dataResult: Result[OutputData[A]]
@@ -77,7 +77,9 @@ trait OutputFactory:
     coll: CC[A]
   )(
     f: A => Output[B]
-  )(using BuildFrom[CC[Output[B]], B, To], Context): Output[To] = sequence(coll.map(f).asInstanceOf[CC[Output[B]]])
+  )(using BuildFrom[CC[Output[B]], B, To], Context): Output[To] = sequence(
+    coll.iterator.map(f).asInstanceOf[CC[Output[B]]]
+  )
   def fail(t: Throwable)(using Context): Output[Nothing] = Output.fail(t)
 trait OutputExtensionsFactory:
   implicit final class OutputSequenceOps[A, CC[X] <: IterableOnce[X], To](coll: CC[Output[A]]):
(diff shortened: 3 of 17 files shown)


core/src/main/scala/besom/internal/Output.scala
```diff
@@ -77,15 +77,22 @@ trait OutputFactory:
...
   def fail(t: Throwable)(using Context): Output[Nothing] = Output.fail(t)
 trait OutputExtensionsFactory:
   implicit final class OutputSequenceOps[A, CC[X] <: IterableOnce[X], To](coll: CC[Output[A]]):
     def sequence(using BuildFrom[CC[Output[A]], A, To], Context): Output[To] =
       Output.sequence(coll)
   implicit final class OutputTraverseOps[A, CC[X] <: IterableOnce[X]](coll: CC[A]):
     def traverse[B, To](f: A => Output[B])(using BuildFrom[CC[Output[B]], B, To], Context): Output[To] =
-      coll.map(f).asInstanceOf[CC[Output[B]]].sequence
+      coll.iterator.map(f).asInstanceOf[CC[Output[B]]].sequence
+  implicit final class OutputOptionOps[A](output: Output[Option[A]]):
```
pawelprazak, line 91: @lbialy please have a look if those two methods make sense in this context

prolativ: Is there any specific reason why implicit classes are used here instead of scala 3 extension methods?

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -77,15 +77,22 @@ trait OutputFactory:
...
+    def getOrElse[B >: A](default: => B)(using Context): Output[B] =
```
prolativ, line 92: IMO it would be nice if a user could also pass values wrapped in `Output` as an argument as well
```suggestion
    def getOrElse[B >: A](default: => B | Output[B])(using Context): Output[B] =
```

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -77,15 +77,22 @@ trait OutputFactory:
...
+      output.flatMap(_.fold(Output(default))(Output(_)))
+    def orElse[B >: A](alternative: => Option[B]): Output[Option[B]] =
```
prolativ, line 94: Maybe
```suggestion
    def orElse[B >: A](alternative: => Option[B] | Output[Option[B]]): Output[Option[B]] =
```?

File: core/src/main/scala/besom/util/NonEmptyString.scala [hunk 4]
```diff
@@ -3,20 +3,42 @@ package besom.util
...
+
+  /** @param s
+   * a [[String]] to be converted to a [[NonEmptyString]].
+   * @return
+   * a [[NonEmptyString]] if the given [[String]] is not empty or blank.
+   */
   inline def from(inline s: String): NonEmptyString = ${ fromImpl('s) }
 
-  def fromImpl(expr: Expr[String])(using quotes: Quotes): Expr[NonEmptyString] =
+  private def fromImpl(expr: Expr[String])(using quotes: Quotes): Expr[NonEmptyString] =
```
pawelprazak, line 41: I've assumed it was not intentionally exposed.

core/src/main/scala/besom/types.scala
```diff
@@ -91,8 +150,15 @@ object types:
   object Label:
     def fromNameAndType(name: NonEmptyString, rt: ResourceType): Label = s"$name[$rt]"
 
+  /** URN is and automatically constructed globally unique identifier for the resource
```
pawelprazak, line 153:
```suggestion
  /** URN is an automatically constructed globally unique identifier for the resource
```

core/src/main/scala/besom/internal/BesomModule.scala
```diff
@@ -3,40 +3,62 @@ package besom.internal
...
-        _              <- BesomLogger.setupLogger()
-        runInfo        <- RunInfo.fromEnv
-        _              <- logger.debug(s"Besom starting up in ${if runInfo.dryRun then "dry run" else "live"} mode.")
-        taskTracker    <- TaskTracker()
-        monitor        <- Result.resource(Monitor(runInfo.monitorAddress))(_.close())
-        engine         <- Result.resource(Engine(runInfo.engineAddress))(_.close())
-        _              <- logger.debug(s"Established connections to monitor and engine, spawning streaming pulumi logger.")
-        logger         <- Result.resource(BesomLogger(engine, taskTracker))(_.close())
-        config         <- Config.forProject(runInfo.project)
+        _           <- BesomLogger.setupLogger()
```
pawelprazak, line 28: don't know why `scalafmt` decided to change this

core/src/main/scala/besom/internal/Output.scala
pawelprazak: I've just followed the file's style, but actually was surprised as well, CC @lbialy

core/src/main/scala/besom/types.scala
```diff
@@ -68,20 +116,31 @@ object types:
...
+      *   a function token string to parse
+      * @return
+      *   a [[FunctionToken]] if the string is valid, otherwise an compile time error occurs
+      */
     // validate that function token contains two colons between three identifiers, see @ResourceType
     inline def from(s: String): FunctionToken =
       requireConst(s)
-      inline if !constValue[Matches[s.type, ".+:.+:.+"]] then
-        error("Invalid function token")
+      inline if !constValue[Matches[s.type, ".+:.+:.+"]] then error("Invalid function token")
```
pawelprazak, line 135: `scalafmt` insisted on this formating

core/src/main/scala/besom/internal/Output.scala
prolativ: Actually `Option[B] | Output[Option[B]]` could be written more shortly as `Input[Option[B]]`. The question is whether we would like to constrain the usage of `Input` to `Args` classes or not

core/src/main/scala/besom/internal/Output.scala
pawelprazak: reimplemented according to the suggestion

core/src/main/scala/besom/internal/Output.scala
pawelprazak: reimplemented according to the suggestion

core/src/main/scala/besom/internal/BesomModule.scala
lbialy: maxColumns=120, should be 140, set maxColumns=140, reformat it back to how it was, don't check in change in .scalafmt.conf into this commit, just revert the reformatting here, we have to fix this in one go everywhere but let's not break our blame needlessly because we'll mark the reformatting commit as blame-transparent for git later

core/src/main/scala/besom/internal/Output.scala
lbialy: @KacperFKorban had a good reason for `implicit class` here, I think it was related to `import x.*` vs `import x.{*, given}` but that doesn't sound correct in regard to extension methods? Kacper, could you weigh in?

core/src/main/scala/besom/types.scala
lbialy: artifact from maxColumns=100 time I think, it's fine

core/src/main/scala/besom/util/NonEmptyString.scala
lbialy: good assumption

core/src/main/scala/besom/internal/Output.scala
KacperFKorban: This issue is in fact about extension methods.

The thing is that we don't just want to use `extension` methods. That's because extension methods can also be called as regular methods and this litters the scope, since in some cases we only want it to be an extension. That's why a true extension only method is defined as a given with that extension method inside e.g.:
```
given AnyRef with
  extension (i: Int)
    def foo = i + 1
```

This goes back to needing to do `import x.given` in addition to the usual `import x.*`.

lbialy:
Ah, yeah. And an implicit class is given with extension but without given.

core/src/main/scala/besom/future.scala
```diff
@@ -11,5 +14,22 @@ trait FutureMonadModule extends BesomModule:
   implicit val toFutureFuture: Result.ToFuture[Eff] = new Result.ToFuture[Future]:
     def eval[A](fa: => Future[A]): () => Future[A] = () => fa
 
+/**
+ * Besom API entry point for the [[scala.concurrent.Future]] monad.
+ *
+ * Please see `run` and `exports` methods to get started.
+ * Other notable methods are `log` and `config`.
```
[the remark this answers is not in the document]
pawelprazak: done

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -77,15 +77,26 @@ trait OutputFactory:
...
     def sequence(using BuildFrom[CC[Output[A]], A, To], Context): Output[To] =
       Output.sequence(coll)
   implicit final class OutputTraverseOps[A, CC[X] <: IterableOnce[X]](coll: CC[A]):
     def traverse[B, To](f: A => Output[B])(using BuildFrom[CC[Output[B]], B, To], Context): Output[To] =
-      coll.map(f).asInstanceOf[CC[Output[B]]].sequence
+      coll.iterator.map(f).asInstanceOf[CC[Output[B]]].sequence
+  implicit final class OutputOptionOps[A](output: Output[Option[A]]):
+    def getOrElse[B >: A](default: => B | Output[B]): Output[B] =
+      default match
+        case b: Output[B] => for o <- output; b <- b yield o.getOrElse(b)
```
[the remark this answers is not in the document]
pawelprazak: done

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -77,15 +77,26 @@ trait OutputFactory:
...
+        case b: B         => for o <- output yield o.getOrElse(b)
+    def orElse[B >: A](alternative: => Option[B] | Output[Option[B]]): Output[Option[B]] =
+      alternative match
+        case a: Output[Option[B]] => for o <- output; a <- a yield o.orElse(a)
+        case a: Option[B]         => for o <- output yield o.orElse(a)
```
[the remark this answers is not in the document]
pawelprazak: done

pawelprazak:
addressed the comments, added `ConfigError` and cleaned up Config factory methods from `throws`

core/src/main/scala/besom/future.scala
```diff
@@ -3,13 +3,40 @@ package besom
...
 
+/** Besom API entry point for the [[scala.concurrent.Future]] monad.
+  *
+  * All Pulumi programs are executed in Besom context [[besom.Context]]
+  *
+  * Most notable methods exposed by [[besom.Pulumi]] are:
+  *   - [[besom.internal.BesomModule.run]] - the Pulumi program function
+  *   - [[besom.internal.BesomSyntax.exports]] - the Pulumi Stack outputs
+  *   - [[besom.internal.BesomSyntax.config]] - configuration and secrets
+  *   - [[besom.internal.BesomSyntax.log]] - all you logging needs
```
lbialy, line 24: typo xD `you` -> `your`

core/src/main/scala/besom/internal/BesomModule.scala
```diff
@@ -3,30 +3,63 @@ package besom.internal
...
         featureSupport <- FeatureSupport(monitor)
-        _              <- logger.trace(s"Environment:\n${sys.env.toSeq.sortBy(_._1).map((k, v) => s"$k: $v").mkString("\n")}")
-        _              <- logger.debug(s"Resolved feature support, spawning context and executing user program.")
-        ctx            <- Context(runInfo, taskTracker, monitor, engine, logger, featureSupport, config)
-        userOutputs    <- program(using ctx).map(_.toResult).getValueOrElse(Result.pure(Struct()))
-        _              <- Stack.registerStackOutputs(runInfo, userOutputs)(using ctx)
-        _              <- ctx.waitForAllTasks
+        _   <- logger.trace(s"Environment:\n${sys.env.toSeq.sortBy(_._1).map((k, v) => s"$k: $v").mkString("\n")}")
+        _   <- logger.debug(s"Resolved feature support, spawning context and executing user program.")
+        ctx <- Context(runInfo, taskTracker, monitor, engine, logger, featureSupport, config)
```
lbialy, line 59: why are these 3 lines indented this way? could you set maxColumns to 140 and reformat? I know it's asinine of me to demand ascii graphics in code but I really want to avoid dumb changes in the single reformat commit.

lbialy (changes requested):

core/src/main/scala/besom/future.scala
pawelprazak:
```suggestion
  *   - [[besom.internal.BesomSyntax.log]] - all your logging needs
```

lbialy (approved):

Issue #295, opened 2023-11-14
Provide `pulumiResourceName` and `pulumiProject`, `pulumiOrganization`, `pulumiStack`


dubinsky:
In Pulumi, name given to a resource at construction time can be recovered by calling `pulumiResourceName` method on the `com.pulumi.resources.Resource`. What is the Besom way of doing this?

Thanks!

lbialy:
Hi!

We don't really have `pulumiResourceName` as such (maybe we should?) but it's still possible to access the resource's logical name via URN. Each and every resource has to have a `urn` field of type `Output[URN]` where URN is an opaque type with several extensions allowing access to it's parts (one of which is a logical resource name!). Therefore:

```scala
    val redisNamespace = Namespace(s"redis-cluster-namespace-$name")

    redisNamespace.urn.map(_.resourceName): Output[String]
```

I think we can add some extension syntaxes for all resources to just have this and that wouldn't be a breaking change. Adding a field to each generated resource to hold on to the logical name would be a breaking change but maybe we should make it. 

Do you feel that getting that name in `Output[String]` is inconvenient?

dubinsky:
> I think we can add some extension syntaxes for all resources to just have this and that wouldn't be a breaking change. Adding a field to each generated resource to hold on to the logical name would be a breaking change but maybe we should make it.

Adding to `besom.internal.Resource` `def principalResourceName: Output[String] = resource.flatMap(_.urn).map(_.resourceName)` could be nice, but even without it everything works great. Thanks!

> Do you feel that getting that name in Output[String] is inconvenient?

It's perfectly fine!
(And will only get better with https://github.com/VirtusLab/besom/issues/291...)

Thank you!!

pawelprazak:
@dubinsky I'm glad you're having fun with Besom, please let us know if any other issue comes to mind :)

@lbialy maybe we could improve the API a bit and it is still less characters than `pulumiResourceName` ;)
`redisNamespace.urn.map(_.resourceName)` -> `redisNamespace.urn.resourceName`

dubinsky:
> I'm glad you're having fun with Besom, please let us know if any other issue comes to mind :)

I am having lots of fun! Thank you!!

1) I need things like `OrganizationsFunctions.getOrganization(GetOrganizationArgs)` (GCP), but I see that they are in the next release. Thanks! When is the next release planned?
[...]

PR #297, opened 2023-11-14
Update docs website

pawelprazak:
A small website update inspired by #295


lbialy (changes requested):
I'd hold my horses with this and implement extension methods for `Output[Resource]` before, it's few lines of code but will require changes here.

pawelprazak:
> I'd hold my horses with this and implement extension methods for `Output[Resource]` before, it's few lines of code but will require changes here.

On the other hand, we'll have to update the docs, after the release, in a lots of places anyway, so we can do it incrementally.

pawelprazak:
blocked on @lbialy

lbialy:
ok I guess

lbialy (approved):

PR #299, opened 2023-11-14
Expose Pulumi organisation information

pawelprazak:
Ported from Go SDK
https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumi/context.go#L221

The organization in Stack is explained in couple places:
- https://www.pulumi.com/docs/concepts/stack/#create-stack
- https://www.pulumi.com/docs/concepts/stack/#stackreferences

Alternatively, to not pollute the context, we could make this a part of stack name, like we did with the URN, WDYT?


pawelprazak:
Change scope and shape of this change according to #295

pawelprazak:
closed in favor of #301

PR #310, opened 2023-11-28
Invoke grpc call 

lbialy:
implemented invoke, added `-Werror -Wunused:all -deprecation -feature` and cleaned up the core codebase


lbialy:
codegen tests fail for obvious reasons, needs work on codegen side of the equation

pawelprazak (approved):
LGTM

PR #312, opened 2023-11-29
Fix provider inputs serialization

pawelprazak:


core/src/main/scala/besom/internal/ProtobufUtil.scala
```diff
@@ -7,7 +7,9 @@ import scala.util.*
 object ProtobufUtil:
   private val printer = JsonFormat.printer().omittingInsignificantWhitespace()
 
-  extension (v: Value) def asJsonString: Either[Throwable, String] = Try(printer.print(Value.toJavaProto(v))).toEither
+  extension (v: Value)
+    def asJsonString: Either[Throwable, String] = Try(printer.print(Value.toJavaProto(v))).toEither
+    def asJsonEncoded: Try[Value]               = v.asJsonString.map(_.asValue).toTry
```
lbialy, line 12: this seems to be dead code in the end, remove pls

core/src/test/scala/besom/internal/ArgsEncoderTest.scala
```diff
@@ -0,0 +1,66 @@
...
+        suppressHelmHookWarnings = suppressHelmHookWarnings.asOptionOutput(isSecret = false)
+      )
+end ArgsEncoderTest
+
+class ArgsEncoderTest extends munit.FunSuite with ValueAssertions:
+  import ArgsEncoderTest.*
+
+  test("encode ProviderArgs") {
+    val pae       = summon[ProviderArgsEncoder[ProviderArgs]]
+    given Context = DummyContext().unsafeRunSync()
```
lbialy, line 57: pardon, my bad, this context is unused here, right? can it be deleted?

core/src/test/scala/besom/internal/PropertiesSerializer.test.scala
```diff
@@ -0,0 +1,52 @@
+package besom.internal
```
lbialy, line 1: name of this file is redundant, no need for .test.scala, it's in directory test

lbialy (changes requested):
all of the provider/argsencoder tests are spluttered in different files, could you move them to a single file or just place in respective test suites, currently they are in:
* ContextTest (probably legacy naming, it doesn't seem to test context and there's not much to test there anyway)
* ArgsEncoderTest
* ProviderArgsEncoderTest

core/src/test/scala/besom/internal/ArgsEncoderTest.scala
pawelprazak: it is used, I've checked

PR #326, opened 2023-12-04
Downgrade exception to logged error for id/urn property

pawelprazak:
- workaround https://github.com/pulumi/pulumi-kubernetes/issues/2683


lbialy (approved):
great!

PR #332, opened 2023-12-14
Update codegen and package generator

### LINKED ISSUE #144: Refer to other resources by reference instead of ID
TypeScript:

```typescript
export interface MethodSettingsArgs {
...
  /**
     * ID of the REST API
     */
    restApi: pulumi.Input<string | RestApi>;
...
}
```

([another example](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/apigateway/authorizer.ts#L229C46-L229C46))

What we generate currently:

```
final case class MethodSettingsArgs private(
...
  restApi: Output[String],
...
)
```

Relevant schema:
```json
          "restApi": {
            "type": "string",
            "oneOf": [
              {
                "type": "string"
              },
              {
                "type": "string",
                "$ref": "#/types/aws:apigateway%2FrestApi:RestApi"
              }
            ],
            "description": "ID of the REST API\n",
            "willReplaceOnChanges": true
          },
```

### LINKED ISSUE #259: We are missing Config in codegen
The `config` field from Pulumi schema is ignored currently by our codegen, example from upstream:
https://github.com/pulumi/pulumi-kubernetes/blob/master/sdk/java/src/main/java/com/pulumi/kubernetes/Config.java

pawelprazak:
- add codegen for provider config (fixes #259)
- add resource references (fixes #144)
- fix dots in package segments in codegen
- fix deserialization of empty ObjectTypeDefinitionOrTypeReference
- resolve dependencies during package generation and publish
- disable bloop for packages
- use 32GB heap for package compilation
- uncomment two fixed codegen integration tests

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -432,15 +432,15 @@ liftoff:
 	#!/usr/bin/env sh
 	export PULUMI_CONFIG_PASSPHRASE=""
 	cd experimental
-	pulumi --non-interactive --logtostderrup --stack liftoff -y
+	pulumi --non-interactive --logtostderr up--stack liftoff -y
 
 # Reverts the deployment of experimental sample kubernetes Pulumi app from ./experimental directory
 destroy-liftoff:
 	#!/usr/bin/env sh
 	cd experimental
-	if (pulumi --non-interactive --logtostderrstack ls | grep liftoff > /dev/null); then
+	if (pulumi --non-interactive --logtostderr stack ls | grep liftoff > /dev/null); then
 		export PULUMI_CONFIG_PASSPHRASE=""
-		pulumi  --non-interactive --logtostderrdestroy --stack liftoff -y
+		pulumi  --non-interactive --logtostderr destroy --stack liftoff -y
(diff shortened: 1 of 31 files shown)


Justfile
```diff
@@ -427,15 +432,15 @@ liftoff:
 	#!/usr/bin/env sh
 	export PULUMI_CONFIG_PASSPHRASE=""
 	cd experimental
-	pulumi --non-interactive --logtostderrup --stack liftoff -y
+	pulumi --non-interactive --logtostderr up--stack liftoff -y
```
pawelprazak, line 435:
```suggestion
	pulumi --non-interactive --logtostderr up --stack liftoff -y
```

lbialy (approved):
LGTM in regard to not throwing. Shared reservations about naive unionDecoders in private (basically that if A is a subset of fields of B it will parse correctly and win although the correct choice was B, has to be solved with a heuristic in codegen to always sort from types with the most fields to least fields).

Issue #77, opened 2023-06-04
[Epic] Clean up TODOs in Codegen


lbialy:
Review and address the existing TODOs in Codegen.

pawelprazak:
TODOs:
- [ ] revert to throwing when upstream resolves https://github.com/pulumi/pulumi/issues/15024
- [ ] implement docs - scaladoc generation
- [ ] truncated file names with a suffix might still potentially clash with each other
- [ ] Handle allowedPackageNames (#342)
- [ ] Handle stateInputs 
- [ ] Handle aliases (#44)
- [ ] Find some workaround to enable passing the remaining arguments (>253)
- [ ] we might need to do proper URL un-escaping i TypeMapper
- [ ] when translating from schema to Scala, we should refactor to use proper ADTs (might be related to PCL)

PR #353, opened 2024-01-26
Cleanup codegen TODOs

pawelprazak:
I've updated our codegen TODOs, alse added comment in #77


pawelprazak:
self-approving as an FIY, since this is just internal comments change

PR #354, opened 2024-01-26
Add language plugin executors form gradle and maven

### LINKED ISSUE #303: Scala Language Plugin for Pulumi should support all the executors that Java Language Plugin does
[Getting started](https://virtuslab.github.io/besom/docs/getting_started/) creates an impression that Scala CLI and Scala Language Plugin are required to use Besom. In fact, if the project uses one of the execution methods supported by the Pulumi Java Language Plugin (Gradle, Maven, JAR, JBang or SBT), Besom SDK can be used with `runtime: java` in `Pulumi.yaml`, just as Pulumi Java SDK can be used from Scala code that way. Indeed, if the project uses Gradle, Maven or JBang, Scala Language Plugin must **not** be used: while retaining support for JAR and SBT execution, it drops support for Gradle etc. in favour of Scala CLI. Scala Language Plugin seems to be required only if the project uses Scala CLI, and although this setup is recommended, I think the ability to use Besom with - say - Gradle should be at least mentioned ;)

pawelprazak:
Fixes #303


integration-tests/integration.scala
```diff
@@ -14,9 +14,9 @@ val javaVersion                 = Config.DefaultJavaVersion
 val scalaVersion                = Config.DefaultScalaVersion
 val coreVersion                 = os.read(os.pwd / "version.txt").trim
 val scalaPluginVersion          = coreVersion
-val providerRandomSchemaVersion = "4.14.0"
-val providerTlsSchemaVersion    = "4.11.1"
-val providerPurrlSchemaVersion  = "0.4.1"
+val providerRandomSchemaVersion = "4.15.1"
+val providerTlsSchemaVersion    = "5.0.0"
+val providerPurrlSchemaVersion  = "0.5.0"
```
lbialy, line 19: we're forced to bump this every time now?

integration-tests/resources/executors/gradle/.gitattributes
```diff
@@ -0,0 +1,9 @@
+#
+# https://help.github.com/articles/dealing-with-line-endings/
+#
+# Linux start script should use lf
+/gradlew        text eol=lf
+
+# These are Windows script files and should use crlf
+*.bat           text eol=crlf
+*.cmd           text eol=crlf
```
lbialy, line 9: I'm anxious to ask what horrible issues with gradle forced this.

lbialy (approved):
lgtm

integration-tests/resources/executors/gradle/.gitattributes
pawelprazak: gradle and mvn have wrappers that are helpful to stabilize the build, bu default they come with windows support, and bat/cmd uses CRLF, it's a git problem mostly

integration-tests/integration.scala
pawelprazak: yes, I'll reuse `scripts/Version` in the future, so version update will be automatic

Issue #307, opened 2023-11-22
Redesign ResourceOptions family


lbialy:
This issue captures all of the issues that are present with current design of ResourceOptions. CRO is currently quite... unwieldy. This structure is used extremely often and yet it requires user to type CustomResourceOptions or ComponentResourceOptions to set a single field in it. This is bad DX. To make things worse arguments on most RO are:
* broken (like Custom Timeouts for instance, the type used - `String` - does not represent them at all and does not work in runtime
* have a non-flexible argument type (for instance dependsOn takes an Output[List[Resource]] which forces user to `.map(List(_))` all single resource cases and also `.zip` to do the same with multiple resources)
* use old argument syntax with `| NotProvided` union whereas codegenerated libs have all moved to `Input[A]`

The inspiration to improve this comes, surprisingly, from the pulumi-go sdk. In pulumi-go user does not create a struct to pass it to the resource constructor. On the contrary, the DX is quite interesting (it's definitely not without issues!) - resource constructor, after the resource name argument and args argument expects any amount of ResourceOption via a variadic argument. Here are some examples:

```go
db, err := NewDatabase(ctx, "db", &DatabaseArgs{ /*...*/ },
    pulumi.AdditionalSecretOutputs([]string{"password"}))
```

```go
res1, _ := NewMyResource(ctx, "res1", &MyResourceArgs{/*...*/})
res2, _ := NewMyResource(ctx, "res2", &MyResourceArgs{/*...*/}, pulumi.DependsOn([]pulumi.Resource{res1}))
```

Each of these functions creates an instance of ResourceOption (notice the lack of plural s!) that is later folded onto the struct that is instantiatated in pulumi-go code and not exposed to the user. This is quite interesting and definitely improves otherwise tricky go DX. To compare this with python, which Scala's SDK is probably closest syntactically to:

```python
db = Database('db',
    opts=ResourceOptions(additional_secret_outputs=['password']))
```

```python
res1 = MyResource("res1")
res2 = MyResource("res2", opts=ResourceOptions(depends_on=[res1]))
```

I'm not exactly sure if variadic argument solution is exactly what we should go towards here mostly because these options would then need to pollute our top-level import scope. For instance for resources that have optional Args like Namespace:
```scala
val db = Namespace("db",  opts = additionalSecretOutputs("a", "b"), dependsOn(k8sProvider, someOtherResource))
```
[...]

Issue #284, opened 2023-11-10
Add support for Component resources in codegen


pawelprazak:
Add support for Component resources in codegen (`isComponent`)

Related #43

PR #355, opened 2024-01-28
Overhauled ResourceOptions and implemented Remote Components

lbialy:
as in title


core/src/main/scala/besom/internal/Input.scala
```diff
@@ -53,6 +54,13 @@ object Input:
       val outputList = inputList.map(simpleInputOps.asOutput(_)(isSecret = isSecret))
       Output.sequence(outputList)
 
+    extension [A](input: Input.OneOrList[A])
```
pawelprazak, line 57: ❤️

pawelprazak (approved):
thank you, looks great!

lbialy:
closes #307 and #284

PR #358, opened 2024-01-30
Add AWS EKS example

pawelprazak:


pawelprazak:
the example hangs when run with #355, reason unknown

lbialy (approved):

PR #360, opened 2024-01-31
Fix traverse on Output

lbialy:
uh, Iterable instead of IterableOnce, thanks Scala Collections!


pawelprazak (approved):
I've added a regression test and I've tested manually, the aws example, now passes

PR #361, opened 2024-01-31
Fix fatals hanging the app with FutureRuntime

lbialy:
this crashes the app if a fatal error is encountered (instead of hanging as Future does) because NonFatal().


pawelprazak (approved):
LGTM, I'll re-test manually on eks example

PR #366, opened 2024-02-01
Drop patch from package version suffix

pawelprazak:
Drop patch from package version suffix

- A.B.C-core.X.Y.Z -> A.B.C-core.X.Y


codegen/src/model/SemanticVersion.test.scala
```diff
@@ -0,0 +1,32 @@
...
+  )
+  Map(
+    SemanticVersion(0, 1, 1, "SNAPSHOT") -> "0.1.1-SNAPSHOT",
+    SemanticVersion(1, 1, 0, "SNAPSHOT") -> "1.1-SNAPSHOT",
+    SemanticVersion(1, 0, 0, "SNAPSHOT", "deadbeef") -> "1-SNAPSHOT+deadbeef",
+  ).foreachEntry((input, expected) =>
+    test(s"short format $input") {
+      assertEquals(input.toShortString, expected)
+    }
+  )
```
lbialy, line 32: lacks test of the comparing logic, doesn't it?

scripts/Version.scala
```diff
@@ -80,10 +88,13 @@ object Version:
...
         val latestPackages = latestPackageVersions()
         projectFiles()
-          .collect { case (path, content) if content.linesIterator.exists(besomDependencyPattern.matches) => path -> content }
+          .collect {
+            case (path, content) if content.linesIterator.exists(besomDependencyPattern.matches) => path -> content 
+          }
           .foreachEntry { case (path, content) =>
             val newContent = content.linesIterator
               .map {
+                case line if line.contains("besom-fake-") => line // ignore
```
lbialy, line 97: huh what :D

lbialy (changes requested):
smol things

codegen/src/model/SemanticVersion.test.scala
pawelprazak: done

pawelprazak:
addressed the sem ver bug, merging

PR #371, opened 2024-02-06
fix transitive dependency resolution algorithm

lbialy:
as in the title, we have had our impl from Java impl which was ported from C# impl which is broken and includes custom resource children of any other resources, including local, remote components and other custom resources, this works and is covered with an extensive test

### DIFF
--- a/core/src/main/scala/besom/internal/ResourceOps.scala
+++ b/core/src/main/scala/besom/internal/ResourceOps.scala
@@ -96,7 +96,7 @@ class ResourceOps(using ctx: Context, mdc: BesomMDC[Label]):
   ): Output[R] =
     ???
 
-  private def executeInvoke(tok: FunctionToken, invokeArgs: SerializationResult, opts: InvokeOptions): Result[Value] =
+  private[internal] def executeInvoke(tok: FunctionToken, invokeArgs: SerializationResult, opts: InvokeOptions): Result[Value] =
     def buildInvokeRequest(args: Struct, provider: Option[String], version: Option[String]): Result[ResourceInvokeRequest] =
       Result {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/besom/internal/ResourceOps.scala
```diff
@@ -220,40 +220,67 @@ class ResourceOps(using ctx: Context, mdc: BesomMDC[Label]):
...
+    def findReachableResources(roots: Set[Resource], acc: Set[Resource]): Result[Set[Resource]] =
+      if roots.isEmpty then Result.pure(acc)
       else
-        val current = left.head
+        val current = roots.head
+
         val childrenResult = current match
-          case cr: ComponentResource => ctx.resources.getStateFor(cr).map(_.children)
-          case _                     => Result.pure(Set.empty)
+          case cb: ComponentBase => resources.getStateFor(cb).map(_.children.filterNot(_ == current))
```
pawelprazak, line 262: nice!

core/src/main/scala/besom/internal/ResourceOps.scala
```diff
@@ -220,40 +220,67 @@ class ResourceOps(using ctx: Context, mdc: BesomMDC[Label]):
...
+          case _                 => Result.pure(Set.empty)
 
         childrenResult.flatMap { children =>
-          findAllReachableResources((left - current) ++ children, acc + current)
+          val updatedAcc =
+            current match
+              case dr: DependencyResource => // Dependency resources are added directly to the set, they don't have a state.
```
pawelprazak, line 268: great!

pawelprazak (approved):
Thank you! Looks great :D

PR #376, opened 2024-02-08
Enable scalafmt in CI

pawelprazak:
as in the title


lbialy (approved):
letssss gooooo

pawelprazak (commented):
I've glanced through the changes, they look ok.

PR #377, opened 2024-02-08
Release/v0.2.0

pawelprazak:


pawelprazak:
restored the branch because in this case we want it to stay

PR #385, opened 2024-02-10
Fix URL validation to allow for kubernetes types

### LINKED ISSUE #384: Kubernetes Ingress - URN parser error
Screenshot 2024-02-10 at 16 13 53

pawelprazak:
- allow for a dot in the URN's module part

fixes #384

### DIFF
--- a/core/src/main/scala/besom/types.scala
+++ b/core/src/main/scala/besom/types.scala
@@ -181,7 +181,7 @@ object types:
       *
       * type       = package ":" [ module ":" ] type name ;
       * package    = identifier ;
-      * module     = identifier ;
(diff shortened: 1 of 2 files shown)


lbialy (approved):
I'm wondering how many exceptions to the schema were going to find. We have to release 0.2.1 on Mon.

pawelprazak:
We have a good indication from the codegen. Not that much.

A mitigation strategy would be to unify the Pulumi model in a common module, for things like type tokens and urn's, but we'd have to carefully think it through first. The benefit would be that we already know from codegen what kind of exceptions are there in all schemas in the registry. The downside would be the coupling and more moving parts.

PR #389, opened 2024-02-12
Loosen up and fix URN parsing

### LINKED ISSUE #387: Is colon allowed in resourceName?
As a part of https://github.com/VirtusLab/besom/pull/328, `besom.URN.UrnRegex` changed, including the change from `(?<resourceName>.+)` to `(?<resourceName>[^:]+|[^:]*::[^:]*)`.

With this change, URNs that contain a single colon `:` in the `resourceName` (e.g., `group:info`) became illegal, although the comment in the code suggests that they should remain legal:
```
Stack, Project, and Name Regex: These are strings that do not contain ::.
```
Was this change intentional?

Thanks!

lbialy:
Fixes #387 by allowing more permissive parsing of URNs coming from engine and adhering to 'string' spec in regard to it being any string without double colons.

### DIFF
--- a/core/src/main/scala/besom/types.scala
+++ b/core/src/main/scala/besom/types.scala
@@ -171,46 +171,49 @@ object types:
       * ```
       * urn = "urn:pulumi:" stack "::" project "::" qualified type name "::" name ;
       *
-      * stack   = string ;
-      * project = string ;
-      * name    = string ;
-      * string  = (* any sequence of unicode code points that does not contain "::" *) ;
+      * stack     = string ;
+      * project   = string ;
+      * name      = anystring ; // lol: https://github.com/pulumi/pulumi/commit/516979770f11d3a426239429731cf9f327c38836
+      * string    = (* any sequence of unicode code points that does not contain "::" *) ;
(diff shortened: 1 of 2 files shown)


dubinsky:
@lbialy 
> adhering to 'string' spec in regard to it being any string without double colons.

Not that I need them, but what is wrong with _double_ colons in the resourceName? ;)

lbialy:
It turns out that for URNs not to match spec is not a big deal for Pulumi
as they basically turned resource name into `(?s:.*)`, so yes, it seems
that they did not validate resource names so hard that now they have to
support newlines in them. We could mirror that (and we do, for stuff coming
from engine) but I would like to preserve a bit of sanity in resource names
mostly because I'm convinced that:
a) somebody will or already did implement a parser that partitions by ::
and additional out of spec double colons will trip it
b) newlines in resource names are not escaped in cli and will break tui

Also, ocd. There's a spec. Let's be predictable and respect the spec.

pawelprazak (approved):
LGTM, thank you! Let's see what more will come up in the future ;)

PR #395, opened 2024-02-19
add kubernetes guestbook example

polkx:


pawelprazak:
For the record, this is the upstream example: https://github.com/pulumi/examples/tree/master/kubernetes-ts-guestbook

examples/kubernetes-guestbook/.gitignore
```diff
@@ -0,0 +1,8 @@
+### Scala an JVM
+*.class
+*.log
+../components/.bsp
+components/.scala-build
```
pawelprazak, line 5:
```suggestion
.bsp
.scala-build
```

pawelprazak: gitignore is recursive, so this should work for both simple and components directories

examples/kubernetes-guestbook/simple/Main.scala
```diff
@@ -0,0 +1,166 @@
...
+    )
+  )
+
+  val frontendService = Service(
+    "frontend",
+    ServiceArgs(
+      metadata = ObjectMetaArgs(labels = frontendLabels, name = "frontend"),
+      spec = ServiceSpecArgs(
+        `type` = isMinikube.map(if _ then ServiceSpecType.ClusterIP else ServiceSpecType.LoadBalancer),
+        ports = List(ServicePortArgs(port = 80, targetPort = 80)),
```
pawelprazak, line 151: the `targetPort` and `containerPort` are logically refering to the same phisical port number, so it would be nice if we create a variable that makes sure they stay in-sync after any changes

examples/kubernetes-guestbook/components/Main.scala
```diff
@@ -0,0 +1,131 @@
...
+  image: String,
+  replicas: Int = 1,
+  resources: Option[ResourceRequirementsArgs] = None,
+  ports: List[Int] = List.empty,
+  allocateIPAddress: Option[Boolean] = None,
+  serviceType: Option[ServiceSpecType] = None,
+  env: List[EnvVarArgs] = List.empty
+)
+
+case class ServiceDeployment(ipAddress: Output[Option[String]])(using ComponentBase) extends ComponentResource
```
pawelprazak, line 21: could we check if changing `Output[Option[String]]` to `Input.Optional[String]` leads to better DX? We shouldn't need to wrap with Some at call site then

pawelprazak (approved):
Thank you! Looks good, found a couple of nits only.

examples/kubernetes-guestbook/components/project.scala
```diff
@@ -0,0 +1,5 @@
+//> using scala "3.3.1"
+//> using options -Werror -Wunused:all -Wvalue-discard -Wnonunit-statement
+//> using plugin "org.virtuslab::besom-compiler-plugin:0.2.1"
+//> using dep "org.virtuslab::besom-core:0.2.1"
+//> using dep "org.virtuslab::besom-kubernetes:4.7.1-core.0.2"
```
pawelprazak, line 5: we might need to add for smoke tests to pass:
```
//> using exclude "simple/*"
```

examples/kubernetes-guestbook/simple/project.scala
```diff
@@ -0,0 +1,5 @@
+//> using scala "3.3.1"
+//> using options -Werror -Wunused:all -Wvalue-discard -Wnonunit-statement
+//> using plugin "org.virtuslab::besom-compiler-plugin:0.2.1"
+//> using dep "org.virtuslab::besom-core:0.2.1"
+//> using dep "org.virtuslab::besom-kubernetes:4.7.1-core.0.2"
```
pawelprazak, line 5: we might need to add for smoke tests to pass:
```
//> using exclude "components/*"
```

PR #400, opened 2024-02-21
Fix issue 397

lbialy:
that was a dumb mistake really, copilot / copy-paste based prolly

Public / user-facing component API does not take any *Args so to reuse the internal functions it uses EmptyArgs with it's ArgsEncoder instance, this was copy-pasted in the rush of implementing remote components when RemoteComponents do actually take Args. The change is not binary-incompatible, just a patch version release should be enough.


pawelprazak (approved):
Looks great, thank you!

PR #421, opened 2024-03-27
Try to encode component args as outputs directly

pawelprazak:
I'm not sure this will work, but it might ;)

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -460,8 +460,7 @@ clean-slate-liftoff: clean-sdk
 	#!/usr/bin/env sh
 	just publish-local-core
 	just publish-local-compiler-plugin
-	scala-cli run codegen -- named kubernetes 4.2.0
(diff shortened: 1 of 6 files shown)


lbialy:
why do we need this change in the first place? 🤔 this is *only* used by component resources and the serialization of fields via `(String, Value)` is not an accident - this mirrors a serialization of a resource, not a field, resources are *not* serialized in outputs, they are bags of named properties with values *in* outputs. I don't really think this is a valid encoding for components and if it works it's probably due to recursive serde handling on the other side. Can you provide an explanation for this change?

pawelprazak:
It is more of an invitation to a brainstorming than a fully formed idea at thins point.

I wanted to explore a possible way to move the problem with "missing sub-trees" to serialization where it was solved.

Because our Resource constructors return Output, we might need to think outside the box on how to solve it.

pawelprazak:
wontfix

PR #433, opened 2024-04-03
Use custom pretty printer in logs

pawelprazak:


core/src/main/scala/besom/internal/ResourceDecoder.scala
```diff
@@ -39,12 +41,14 @@ object ResourceDecoder:
               .decode(value, propertyLabel)
               .map(_.withDependency(resource))
               .tapBoth(
-                err => log.debug(s"failed to extract custom property '$propertyName' from '$value': '$err'"),
-                value => log.debug(s"extracted custom property '$propertyName' from '$value'")
+                err => log.debug(s"failed to extract custom property '$propertyName' from '${printer.render(value)}': '$err'"),
+                _ => log.debug(s"extracted custom property '$propertyName' from '${printer.render(value)}'")
```
lbialy, line 45: Wait, this does something different than before, now it prints the value as it was before decoding. Before it printed the result of decoding, didn't it?

core/src/main/scala/besom/internal/ResourceOps.scala
```diff
@@ -487,7 +487,7 @@ class ResourceOps(using ctx: Context, mdc: BesomMDC[Label]):
             ,
             version = inputs.options.version.getOrElse(""), // protobuf expects empty string and not null
             ignoreChanges = inputs.options.ignoreChanges,
-            acceptSecrets = true, // TODO doing like java does it
+            acceptSecrets = true, // our implementation supports secrets form day one
```
lbialy, line 490: form -> from

core/src/main/scala/besom/util/printer.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  override def treeify(x: Any): Tree =
+    val result = super.treeify(x)
+    result match
+      case Tree.Apply(p, body: Iterator[Tree]) =>
+        Tree.Apply(
+          p,
+          body
+            .filterNot {
+              case Tree.KeyValue("unknownFields", _) => true
```
lbialy, line 20: Great!

lbialy (changes requested):
Smol nitpicks, one dubious change. Overall score: veri nice.

core/src/main/scala/besom/internal/ResourceDecoder.scala
pawelprazak: This error message is not correct, I have a correction in another branch.

The `value` in the `tapBoth` was shadowing the `value` in `map` above.
The decoded value will be full of outputs (the ones you've complained about).

pawelprazak: I've reverted the debug message to display the resulting output data, but changed the variable name to get rid of shadowing.

I've changed the printer to turn `besom.internal.Output@3b751a0c` into `Output(?)` in logs.

lbialy (approved):

Issue #435, opened 2024-04-04
Add `Output.when` convenience method


pawelprazak:
Allow for easy combination of `Output[Boolean]` and `Output[Resource]` to mamy `Output[Option[Resource]]`.

e.g. instead of
```
 val databaseInstance: Output[Option[Output[Server]]] =
    bootzookaConfig.dbEnabled.map(Option.when(_)(postgresDatabase(appName, resourceGroup, bootzookaConfig.sql)))
```

allow for:
```
val databaseInstance: Output[Option[Server]] = Output.when(bootzookaConfig.dbEnabled)(postgresDatabase(appName, resourceGroup, bootzookaConfig.sql)
```

PR #439, opened 2024-04-05
Add `Output.when`

pawelprazak:
fixes https://github.com/VirtusLab/besom/issues/435

### DIFF
--- a/core/src/main/scala/besom/internal/Output.scala
+++ b/core/src/main/scala/besom/internal/Output.scala
@@ -80,6 +80,10 @@ trait OutputFactory:
   )(using BuildFrom[CC[Output[B]], B, To], Context): Output[To] = sequence(coll.map(f).asInstanceOf[CC[Output[B]]])
 
   def fail(t: Throwable)(using Context): Output[Nothing] = Output.fail(t)
+
+  def when[A: Typeable](cond: => Input.Optional[Boolean])(a: => Input.Optional[A])(using ctx: Context): Output[Option[A]] =
+    Output.when(cond)(a)
+
 end OutputFactory
 
 trait OutputExtensionsFactory:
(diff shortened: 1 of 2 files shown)


core/src/main/scala/besom/internal/Output.scala
```diff
@@ -80,6 +80,11 @@ trait OutputFactory:
   )(using BuildFrom[CC[Output[B]], B, To], Context): Output[To] = sequence(coll.map(f).asInstanceOf[CC[Output[B]]])
 
   def fail(t: Throwable)(using Context): Output[Nothing] = Output.fail(t)
+
+  def when[A: Typeable](cond: => Boolean | Output[Boolean] | Output[Option[Boolean]])(
```
lbialy, line 84: `Input.Optional[Boolean]` instead

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -80,6 +80,11 @@ trait OutputFactory:
...
+    a: => A | Output[A] | Output[Option[A]]
```
lbialy, line 85: same

core/src/test/scala/besom/internal/OutputTest.scala
```diff
@@ -138,4 +138,46 @@ class OutputTest extends munit.FunSuite:
...
+    given Context = DummyContext().unsafeRunSync()
+
+    val v: Output[String] = Output({
+      fail("This should not be evaluated")
+      "value"
+    })
+    val result = Output.when(false)(v)
+    assertEquals(result.getData.unsafeRunSync(), OutputData(None))
+  }*/
+
```
lbialy, line 182: uh what?

lbialy (changes requested):
one or two things

core/src/main/scala/besom/internal/Output.scala
pawelprazak: done

core/src/test/scala/besom/internal/OutputTest.scala
pawelprazak: this should not be here, removed, thanks

lbialy (approved):

PR #441, opened 2024-04-08
Fix traverse problem from #430, fix component API function too

### LINKED ISSUE #430: Exception upon traverse
When running the [example](https://github.com/VirtusLab/besom/blob/3cba8863f80f0ab81af245f19dc89423187ec868/examples/aws-s3-folder/Main.scala#L64C14-L64C45 
) , prior removal of type information in the `val uploads = ` declaration leads to a runtime exception  

```
class scala.collection.immutable.$colon$colon cannot be cast to class scala.collection.View
[error] 	at scala.collection.BuildFrom$$anon$8.newBuilder(BuildFrom.scala:85)
[error] 	at besom.internal.Output$.sequence$$anonfun$1$$anonfun$1(Output.scala:123)
```

lbialy:
Fixes #430


pawelprazak (approved):
Looks good, thank you! :)

PR #443, opened 2024-04-08
Improve errors around `Output.eval` and `Output#flatMap` / `ToFuture` `@implicitNotFound`

lbialy:
Adds nice compile error for wrong usage of flatMap and for missing instance of Result.ToFuture


pawelprazak (approved):
Looks great, thank you!

PR #445, opened 2024-04-09
Add lifts to `Output[Option[List[A]]]` etc.

### LINKED ISSUE #424: Consider adding lifted methods for `Output[List[A]]` and `Output[Option[List[A]]]`, and `Output[Option[A]]`
In the following use case:
```
val nginxService = Service(
    "nginx",
    ServiceArgs(
      spec = ServiceSpecArgs(
        selector = labels,
        `type` = k8s.core.v1.enums.ServiceSpecType.LoadBalancer,
        ...
      ),...
  )
  
  Stack.exports(
    nginxUrl = nginxService.status.loadBalancer.ingress.head.hostname
  )
```

what is missing is:
```
  extension [A](output: besom.types.Output[scala.Option[scala.collection.immutable.List[A]]])
    def head: besom.types.Output[scala.Option[A]] = output.map(_.map(_.head))
```

We should consider adding lifts for both `Output[List[A]]` and `Output[Option[List[A]]]`, for the essential methods: i.e.: `head`, `tail`, `get`, etc. (`mapList`?), but adding every single method does not feel practical.

EDIT:

A more complex use case:
```
  extension [A](output: Output[Option[A]])
    def orElse[B >: A](alternative: Output[Option[B]]): Output[Option[B]] =
      output.flatMap(o => alternative.map(a => o.orElse(a)))

  extension [A](output: Output[Option[List[A]]]) 
    def headOption: Output[Option[A]] = output.map(_.flatMap(_.headOption))

  val maybeIngress = nginxService.status.loadBalancer.ingress.headOption
  val hostnameOrIp = maybeIngress.hostname.orElse(maybeIngress.ip)

  Stack.exports(
    nginxHostname = p"http://${hostnameOrIp.getOrElse("unknown")}:$nginxPortNumber"
  )
```

pawelprazak:
Fixes #424

### DIFF
--- a/core/src/main/scala/besom/internal/Output.scala
+++ b/core/src/main/scala/besom/internal/Output.scala
@@ -6,7 +6,7 @@ import scala.reflect.Typeable
 /** Output is a wrapper for a monadic effect used to model async execution that allows Pulumi to track information about dependencies
   * between resources and properties of data (whether it's known or a secret for instance).
   *
-  * Invariant: dataResult has to be registered in TaskTracker by the time it reaches the constructor here!
+  * Invariant: `dataResult` has to be registered in [[TaskTracker]] by the time it reaches the constructor here.
   * @param dataResult
   *   Effect of type [[Result]][A]
   * @param ctx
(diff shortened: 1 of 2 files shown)


pawelprazak:
Added the requested `getOrFail`, and scaladocs for public methods.

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -113,15 +244,162 @@ trait OutputExtensionsFactory:
...
+      */
+    def headOption: Output[Option[A]] = output.map(_.headOption)
+
+    /** Calls [[List.lastOption]] on the underlying [[List]]
+      * @return
+      *   an [[Output]] of [[Option]] of the last element of the list
+      */
+    def lastOption: Output[Option[A]] = output.map(_.lastOption)
+
+    /** Calls [[List.tail]] on the underlying [[List]], but does not throw on `Nil`
```
lbialy, line 317: fail instead of throw

core/src/main/scala/besom/internal/Output.scala
```diff
@@ -113,15 +244,162 @@ trait OutputExtensionsFactory:
...
+      * @return
+      *   an [[Output]] of the `tail` of the [[List]] or an empty list if the list is empty
+      */
+    def tailOrEmpty: Output[List[A]] = output.map {
+      case Nil  => Nil
+      case list => list.tail
+    }
+
+    /** Calls [[List.init]] on the underlying [[List]], but does not throw on `Nil`
```
lbialy, line 326: fail instead of throw

lbialy (changes requested):

pawelprazak:
Addressed the comments. Extended signatures of mapInner / flatMap inner for lists to match option. Added unit tests.

lbialy (approved):

PR #447, opened 2024-04-10
changes to docs for 0.3, first draft

lbialy:
* better explanation of resource constructor syntax
* better explanation of Output/Input dynamics, how they work
* draft of explanation of how to structure besom programs, I think we should expand upon this


website/docs/async.md
```diff
@@ -0,0 +1,119 @@
...
+for 
+  a <- Bucket("first")
+  b <- Bucket("second")
+yield ()
+```
+
+  </TabItem>
+</Tabs>
+
+There is a way to inform Pulumi engine that some of the resources have to be created, updated or deleted
```
pawelprazak, line 105: It would be worth noting that "depends on" is needed only when Pulumi cannot infer the sequence from output dependency graph

pawelprazak (approved):
Thank you, got one suggestion.

website/docs/async.md
lbialy: yes, this is exactly what I had in mind here:

draft of explanation of how to structure besom programs, I think we should expand upon this

but it feels like a big topic and I'm thinking about a best practices section to distill what happens if you do stuff what way there and explain it a bit more

pawelprazak: I'd add a word explicitly after inform.

And add a short remark that Pulumi tries to infer sequence from dependency graph.

PR #452, opened 2024-04-11
Add Azure ACI example

polkx:
Example based on:
https://github.com/pulumi/examples/tree/master/azure-ts-aci

### DIFF
--- a/examples/azure-aci/Main.scala
+++ b/examples/azure-aci/Main.scala
@@ -39,6 +39,6 @@ import besom.api.azurenative
 
   Stack(containerGroup).exports(
     // TODO uncomment when bug https://github.com/VirtusLab/besom/issues/432 will be fixed
-    // containerIPv4Address = containerGroup.ipAddress.ip
+    containerIPv4Address = containerGroup.ipAddress.ip
   )
 }
--- a/examples/azure-aci/Main.scala
+++ b/examples/azure-aci/Main.scala
@@ -1,0 +1,44 @@
+import besom.*
+import besom.api.azurenative
+
+@main def main = Pulumi.run {
+  val resourceGroup = azurenative.resources.ResourceGroup("resourceGroup")
+
+  val imageName = "mcr.microsoft.com/azuredocs/aci-helloworld"
+
+  val containerGroup = azurenative.containerinstance.ContainerGroup(
+    name = "containerGroup",
+    azurenative.containerinstance.ContainerGroupArgs(
+      resourceGroupName = resourceGroup.name,
+      osType = azurenative.containerinstance.enums.OperatingSystemTypes.Linux,
+      containers = List(
+        azurenative.containerinstance.inputs.ContainerArgs(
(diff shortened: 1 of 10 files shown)


pawelprazak:
Thnk you, please rebase on main and update core libs to `0.3.0-SNAPSHOT` and providers to `0.3-SNAPSHOT`.

examples/azure-aci/Main.scala
```diff
@@ -0,0 +1,44 @@
...
+        ),
+        `type` = azurenative.containerinstance.enums.ContainerGroupIpAddressType.Public
+      ),
+      restartPolicy = azurenative.containerinstance.enums.ContainerGroupRestartPolicy.Always
+    )
+  )
+
+  Stack(containerGroup).exports(
+    // TODO uncomment when bug https://github.com/VirtusLab/besom/issues/432 will be fixed
+    // containerIPv4Address = containerGroup.ipAddress.ip
```
pawelprazak, line 42: @lbialy would you be able to provide the Output.recover to enable a workaround here?

pawelprazak (changes requested):

pawelprazak:
@lbialy would you be able to provide Output.recover to enable a workaround here?

examples/azure-aci/Main.scala
lbialy: While I agree that we should add recover, I don't really see wow would recover help here. All it will do is allow user to substitute the broken property with something provided by the user.

pawelprazak: Since the `ip` itself is present and correct, the idea was to workaround this, by ignoring the exception and getting the value.

lbialy (changes requested):
Can you uncomment the export and check this on 0.4.0-SNAPSHOT core? It should work now.

polkx:
bump version to 0.4.0-SNAPSHOT and uncomment export but it is still not working. Same error as before

lbialy:
Ok, I'll take a look at it.

lbialy:
```
λ pulumi up
Previewing update (dev)

     Type                                              Name            Plan
 +   pulumi:pulumi:Stack                               azure-aci-dev   create
 +   ├─ azure-native:resources:ResourceGroup           resourceGroup   create
 +   └─ azure-native:containerinstance:ContainerGroup  containerGroup  create

Outputs:
    containerIPv4Address: output<string>

Resources:
    + 3 to create

Do you want to perform this update? yes
Updating (dev)

     Type                                              Name            Status
 +   pulumi:pulumi:Stack                               azure-aci-dev   created (48s)
 +   ├─ azure-native:resources:ResourceGroup           resourceGroup   created (4s)
 +   └─ azure-native:containerinstance:ContainerGroup  containerGroup  created (38s)

Outputs:
    containerIPv4Address: "<REDACTED>"

Resources:
    + 3 created

Duration: 55s
```
After rebase onto main works fine.

lbialy (approved):

PR #463, opened 2024-04-17
Update mangled properties list

### LINKED ISSUE #458: Fix codegen for `fortios` provider
A provider compilation failed with:
```
[31m[31m-- [E164] Declaration Error: /Users/pprazak/repos/besom/.out/codegen/fortios/0.0.5/src/router/Bgp.scala:19:2 [0m[0m
[31m19 |[0m  asString: besom.types.Output[[35mString[0m],
[31m[31m   |[0m  ^[0m
[31m   |[0merror overriding [33mmethod[0m [35masString[0m in [33mtrait[0m [35mResource[0m of type => besom.internal.Result[Option[String]];
[31m   |[0m  [33mvalue[0m [35masString[0m of type besom.internal.Output[String] needs `override` modifier
1 warning found
1 error found
Compilation failed
```

`Bgp.scala`:
```
final case class Bgp private(
  urn: besom.types.Output[besom.types.URN],
  id: besom.types.Output[besom.types.ResourceId],
  ...
  as: besom.types.Output[Int],
  asString: besom.types.Output[String],  // <<<< needs to be escaped
  ...
) extends besom.CustomResource
```

pawelprazak:
I believe we have to wait for `0.4.0` for this one to be on the safe side since we are changing the core, unless we can prove this is not a braking change for the `0.3.x` line.

Fixes #458

### DIFF
--- a/codegen/src/CodeGen.test.scala
+++ b/codegen/src/CodeGen.test.scala
@@ -847,6 +847,93 @@ class CodeGenTest extends munit.FunSuite {
       tags = Set(
         munit.Ignore
       ) // FIXME: un-ignore when this is fixed: https://github.com/pulumi/pulumi-kubernetes/issues/2683
+    ),
+    Data(
+      name = "Error on urn property",
+      json = """|{
+                |  "name": "mangled-provider",
+                |  "version": "0.0.1",
+                |  "resources": {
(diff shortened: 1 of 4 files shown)


lbialy (approved):
Probably true, should wait for 0.4.0

PR #465, opened 2024-04-18
fix issue 462 - MatchError in Output[String] -> Output[NonEmptyString] inference

lbialy:
pulumi string interpolation inference macro threw MatchError on dynamic Output[String]


pawelprazak (approved):
Looks good, thanks.

Issue #466, opened 2024-04-18
Add missing combinators to Output


lbialy:
* unzip: `Output[(A, B, C)]` -> `(Output[A], Output[B], Output[C])`
* recover
* tap
* tapError
* tapBoth

lbialy:
done

PR #467, opened 2024-04-18
add more combinators to Output - issue #466

lbialy:
some discussion regarding destructuring in for-comprehensions is still to be had @pawelprazak


pawelprazak (approved):
Looks great, thank you.

PR #469, opened 2024-04-19
Update scripts

pawelprazak:
- add maven commands to check published versions in the repository
- skip publishing if the version is already published
- add a force flag
- remove publish-maven-no-deps command that was a workaround for not checking published deps

### DIFF
--- a/scripts/Version.scala
+++ b/scripts/Version.scala
@@ -1,5 +1,6 @@
 package besom.scripts
 
+import besom.codegen.Config
 import besom.model.SemanticVersion
 
 import scala.sys.exit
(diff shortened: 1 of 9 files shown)


scripts/Version.scala
```diff
@@ -1,55 +1,31 @@
...
-        .parseTolerant(besomVersion)
-        .fold(
-          e => throw Exception(s"Invalid besom version: ${besomVersion}", e),
-          _.copy(patch = 0).toShortString
-        )
-      version match
-        case s"$a:$b-core.$_" => s"$a:${packageVersion(a).getOrElse(b)}-core.$coreShortVersion"
-        case s"$a:$_"         => s"$a:$newBesomVersion"
-    end changeVersion
+      """^(//> using (?:test\.)?(?:dep|lib|plugin) +"?\S+::besom-)([^"]+)("?)$""".r
```
lbialy, line 17: is it possible to have a comment explaining this pattern a bit? prolly gpt can do this quite well

scripts/common.scala
```diff
@@ -8,11 +8,34 @@ import java.util.concurrent.atomic.AtomicInteger
...
+        (rest, (v, 1) +: flags)
+      case s"--$name" +: value +: tail =>
+        val (rest, flags) = parseInner(tail, monoFlags)
+        (rest, (name, value) +: flags)
+      case s"-$name" +: value +: tail =>
+        val (rest, flags) = parseInner(tail, monoFlags)
+        (rest, (name, value) +: flags)
+      case head +: tail =>
+        val (rest, flags) = parseInner(tail, monoFlags)
+        (head +: rest, flags)
```
lbialy, line 38: at this point it wouldn't it make sense to use a proper cli parser library? case-app? scopt? scallop?

lbialy (changes requested):

scripts/common.scala
pawelprazak: I didn't want to over-engineer a scripts module :)

scripts/Version.scala
pawelprazak: I've added a comment explaining the intent of the regexp and te highlights. I didn't include a full explanation of the regexp symbol by symbol because I find it redundant.

lbialy (approved):

PR #479, opened 2024-04-25
Add AWS API Gateway V2 with EventBridge example

polkx:
Example based on:
https://github.com/pulumi/examples/tree/master/aws-ts-apigatewayv2-eventbridge


pawelprazak (changes requested):
Looks good, thank you!

Please update to 0.4.0-SNAPSHOT, we'll release this in the next cycle :)

pawelprazak (approved):

Issue #489, opened 2024-05-06
besom.internal.ResourceDecoder fails to parse some escape codes


nikolaiser:
```scala
//> using scala "3.3.3"
//> using options -Werror -Wunused:all -Wvalue-discard -Wnonunit-statement
//> using dep "org.virtuslab::besom-core:0.3.1"
//> using dep "org.virtuslab::besom-command:0.10.0-core.0.3"
import besom.*
import besom.api.command

@main def main = Pulumi.run {

  val run = command.local.Command(
      "test-command",
      command.local.CommandArgs(
          create = """echo -e "normal text here \n\x1b[0;1;31m this text should be bold and red""""
        )
    )

  Stack.exports(
    stdout = run.stdout
  )
}
```

`pulumi up` fails with 

```
 2024.05.07 00:02:40:797 scala-execution-context-global-52 [resource: test-command[command:local:Command]] ERROR besom.internal.ResourceDecoder.resolve:97
        Resolve resource: received error from gRPC call: 'Unknown ansi-escape [0;1;31m at index 21 inside string cannot be
        parsed into an fansi.Str', failing resolution
    2024.05.07 00:02:40:844 main  ERROR besom.internal.BesomModule.run:70
        java.lang.IllegalArgumentException: Unknown ansi-escape [0;1;31m at index 21 inside string cannot be parsed into an
         fansi.Str
```

lbialy:
Thank you for this report! I'm wondering if this is hiding another issue as we don't really use fansi in contexts different than debugging and error handling. I'll investigate.

lbialy:
Ok, so this is a much bigger problem than just fansi having issues with parsing some escape codes. It seems Pulumi engine does something to escape codes and even if I disable trace logging that leads to this exception the besom program does not terminate at all in your sample. I guess something goes horribly wrong on pulumi's side but I'm still researching.

lbialy:
Ha, problem is on our side (in fansi and our pretty printer). I have a solution that works.

lbialy:
fixed in #492

PR #492, opened 2024-05-10
Fix issue 489 - trace level messages get evaluated to send to Pulumi engine

lbialy:
Fansi has some issues with some ansi escape codes as reported in #489. This does not fix the inner issue, we just stop sending trace log events to the engine and that prevents evaluation of the lines that cause the problem. Alternative would be to stop using pprint/fansi combo for debug but that would make debugging experience much worse so no go.

Therefore this is fixed for the end user, weird fansi (?) bug persists on trace level and can / will come back to bite us in debugging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
   build:
     if: "! github.event.pull_request.head.repo.fork"
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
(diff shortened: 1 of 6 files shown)


lbialy:
Changed it so that no trace level messages get sent to pulumi engine. Temporary solution.

lbialy:
Found a solution that actually works. No need for half-assed temporary solutions.

lbialy (commented):
lgtm, bravo me

PR #494, opened 2024-05-15
besom-cfg initial checkin to monorepo

lbialy:


Justfile
```diff
@@ -1,6 +1,7 @@
 # Big idea behind using a Justfile is so that we can have modules like in sbt.
 
 besom-version := `cat version.txt`
+besom-cfg-version := `cat besom-cfg/version.txt`
```
pawelprazak, line 4: I assume there is a reason behind versioning cfg separately, but wouldn't it make sense to align the versions, just for the sake of simplicity?

besom-cfg/k8s/project.scala
```diff
@@ -0,0 +1,23 @@
+//> using scala 3.3.3
+
+//> using dep com.lihaoyi::os-lib::0.9.3
+//> using dep org.virtuslab::besom-cfg:0.2.0-SNAPSHOT
```
pawelprazak, line 4: this will be replaced by current besom version next time you run the version bump script, that's why I was suggesting the version alignment ;)

besom-cfg/k8s/src/main/scala/ConfiguredContainerArgs.scala
```diff
@@ -0,0 +1,172 @@
...
+import besom.cfg.*
+import besom.json.*
+import besom.cfg.containers.*
+import besom.api.kubernetes.core.v1.inputs.*
+
+import scala.util.*
+import scala.quoted.*
+import besom.cfg.k8s.syntax.*
+
+// this is besom-cfg-kubernetes entrypoint
```
pawelprazak, line 14: since this is the entrypoint, I'd strongly advice adding even the most barebones scaladoc :)

Justfile
lbialy: no, it wouldn't, this stuff has and has to have a completely separate lifecycle

besom-cfg/k8s/project.scala
lbialy: so there has to be a way to provide exceptions in the version bump script

besom-cfg/k8s/src/main/scala/ConfiguredContainerArgs.scala
lbialy: I'll strongly consider that once besom-cfg is beyond a prototype

PR #498, opened 2024-05-23
An ETL pipeline with Amazon Redshift and AWS Glue example

polkx:
Example based on:
https://github.com/pulumi/examples/tree/master/aws-ts-redshift-glue-etl


lbialy (approved):
checked, works!

PR #501, opened 2024-05-24
Add Azure Synapse Workspace and Pools example

polkx:
Example based on:
https://github.com/pulumi/examples/tree/master/azure-ts-synapse


pawelprazak (approved):

lbialy (approved):
checked and works fine

PR #502, opened 2024-05-24
fix issue 488

lbialy:
This fixes problem with missing `ResourceState` for components due to `A <: ComponentResource` / `ComponentBase` split (`Resources` keeps instances of `ComponentBase` as keys, user can provide `A <: ComponentResource` as lookup value). Solved by exposing internally the `ComponentBase` dependency in any `A <: ComponentResource` and using it as lookup value in `Resources`.


lbialy (commented):
lgtmyself

PR #524, opened 2024-06-24
Add Output type to fields in generated case classes

polkx:


core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -715,37 +722,37 @@ trait DecoderHelpers:
...
+        case (struct, Constants.SpecialSig.OutputSig) =>
           val innerValue: ValidatedResult[DecodingError, Option[Value]] =
             struct.fields
-              .get(ValueName)
+              .get(Constants.ValueName)
               .validResult
           val isSecret: ValidatedResult[DecodingError, Boolean] =
             struct.fields
-              .get(SecretName)
+              .get(Constants.SecretName)
```
lbialy, line 744: Same as above.

core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -715,37 +722,37 @@ trait DecoderHelpers:
...
               .collectFirst({ case Value(Kind.BoolValue(b), _) => b })
               .getOrElse(false)
               .validResult
 
           val deps: ValidatedResult[DecodingError, Vector[Resource]] =
             struct.fields
-              .get(DependenciesName)
+              .get(Constants.DependenciesName)
```
lbialy, line 751: Same as above.

core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -990,12 +998,12 @@ object Encoder:
...
 
   private def assetWrapper(key: String, value: Value): Value = Map(
-    SpecialSig.Key -> SpecialSig.AssetSig.asValue,
+    Constants.SpecialSig.Key -> Constants.SpecialSig.AssetSig.asValue,
     key -> value
   ).asValue
 
   private def archiveWrapper(key: String, value: Value): Value = Map(
-    SpecialSig.Key -> SpecialSig.ArchiveSig.asValue,
+    Constants.SpecialSig.Key -> Constants.SpecialSig.ArchiveSig.asValue,
```
lbialy, line 1006: Same as above.

core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -715,37 +722,37 @@ trait DecoderHelpers:
   def decodeAsPossibleSecretOrOutput(value: Value, label: Label)(using Context): ValidatedResult[DecodingError, OutputData[Value]] =
     value
       .withSpecialSignature {
-        case (struct, SpecialSig.SecretSig) =>
+        case (struct, Constants.SpecialSig.SecretSig) =>
           val innerValue = struct.fields
-            .get(ValueName)
+            .get(Constants.ValueName)
```
lbialy, line 727: Same as above.

core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -229,6 +229,12 @@ object Decoder extends DecoderInstancesLowPrio1:
       if v.kind.isBoolValue then v.getBoolValue.valid
       else error(s"$label: Expected a boolean, got: '${v.kind}'", label).invalid
 
+  given outputDecoder[A](using inner: Decoder[A]): Decoder[Output[A]] = new Decoder[Output[A]]:
+    override def decode(value: Value, label: Label)(using Context): ValidatedResult[DecodingError, OutputData[Output[A]]] =
+      inner.decode(value, label).map(_.map(Output(_)))
```
lbialy, line 234: This is wrong - it misses the whole point of this change! This works in happy case only and we're not interested in happy case because if anything goes wrong with decoding on this level with this implementation it will crash deserialization of the whole entity instead of this single field like we intended. You should create a redeem method on `ValidatedResult` like this:
```scala
    def redeem[B](fe: NonEmptyVector[E] => B, fa: A => B): ValidatedResult[E, B] =
      result.map {
        case Valid(a) => Valid(fa(a))
        case Invalid(e) => Valid(fe(e))
      }
```
and then handle the invalid case by producing a failed `Output` with `Output.fail` containing an `AggregatedDecodingError`. For happy `Valid` path just wrap in `Output()`.

lbialy (approved):

PR #525, opened 2024-06-24
improvements to besom-cfg 

lbialy:


besom-cfg/lib/src/main/scala/Configured.scala
```diff
@@ -3,6 +3,11 @@ package besom.cfg
 import scala.quoted.*
 import besom.json.*
 import besom.cfg.internal.*
+import besom.util.Validated
+import scala.util.control.NoStackTrace
+import org.checkerframework.checker.units.qual.A
```
lbialy, line 8: ffs Metals!!!!

Justfile
```diff
@@ -314,65 +313,106 @@ publish-language-plugins-all: package-language-plugins-all
 compile-cfg-lib: publish-local-json publish-local-core
 	scala-cli --power compile besom-cfg/lib --suppress-experimental-feature-warning
 
+compile-cfg-containers: publish-local-cfg-lib publish-local-model
+	scala-cli --power compile besom-cfg/containers --suppress-experimental-feature-warning
+
 # Compiles besom-cfg k8s module
-compile-cfg-k8s: publish-local-cfg-lib
+compile-cfg-k8s: publish-local-cfg-lib publish-local-cfg-containers
 	just cli packages local kubernetes:4.17.1
```
prolativ, line 321: Is it intentional to have this version fixed here? What about publishing this module to maven? Are we going to always release besom-kubernetes for the latest k8s version and for 4.17.1? Otherwise this module will be unusable when downloaded from maven central, right?

besom-cfg/containers/.scalafmt.conf
prolativ, line 1: Is this file actually used? It looks that our CI uses the `.scalafmt.conf` file from the project's root to verify formatting throughout the project

besom-cfg/lib/src/main/scala/Configured.scala
prolativ, line 8: I believe this line is still to be removed

besom-cfg/lib/src/main/scala/Configured.scala
```diff
@@ -3,6 +3,11 @@ package besom.cfg
...
+
+final val Version = "0.4.0-SNAPSHOT"
```
prolativ, line 10: Yet another place to remember about when bumping versions 😞 
How about using scala-cli's `BuildInfo` to keep this up to date with what we have in the module's build definition?

besom-cfg/lib/src/main/scala/Configured.scala
```diff
@@ -26,75 +31,84 @@ import besom.cfg.internal.*
...
+        }
       }
-    }
+    end derivedImpl
+  end FromEnv
 end Configured
 
-def resolveConfiguration[A](using c: Configured[A]): A =
-  c.newInstanceFromEnv()
+def resolveConfiguration[A, INPUT: Default](using c: Configured[A, INPUT]): A =
```
prolativ, line 102: Is this supposed to be proper API for users?
Should they then call something like `resolveConfiguration[Foo, Map[String, String]]`?
Are we sure this is going to work well when we mix type inference and implicit resolution with overloading?

Justfile
lbialy: version is indeed fixed here and in project.scala for cfg-k8s, does it matter though? macro-stuff is going away, codegen is going to generate classes for the user without depending explicitly on besom-kubernetes.

besom-cfg/lib/src/main/scala/Configured.scala
lbialy: what the hell, I did kill it

lbialy: ok, pushed fix

besom-cfg/lib/src/main/scala/Configured.scala
lbialy: yeah but no, we're moving away from macros, right?

besom-cfg/lib/src/main/scala/Configured.scala
lbialy: this is a good question and I think this was WIP but since we're killing macros-based approach we can ignore this and figure that out later, no?

prolativ: ok

prolativ (approved):

besom-cfg/lib/src/main/scala/Configured.scala
lbialy: fixed

PR #529, opened 2024-07-03
Use Iterable instead of List

### LINKED ISSUE #431: Use more generic collection interfaces in inputs of user API
Currently we require `List` in most user-facing APIs and as a consequence we force the user to add things like `.map(_.toList)` when the API itself does not return a `List`, e.g.:

```scala
ecs.FargateServiceArgs(
          networkConfiguration = ServiceNetworkConfigurationArgs(
            ...
            securityGroups =
              loadBalancer.defaultSecurityGroup.map(_.map(_.id)).map(_.toList) // <-- unnecessary
          ),
...)
```

PawelGizka:
closes #431 

Tested on aws-eks using both Iterable and List as Input. Additionally tested both Iterable and List as stack outputs.


core/src/main/scala/besom/internal/codecs.scala
```diff
@@ -1074,17 +1075,25 @@ object Encoder:
...
+          joinedMetadata -> values.toList.asValue
         }
 
+  given listEncoder[A](using innerEncoder: Encoder[A]): Encoder[List[A]] = iterableEncoder.contramap[List[A]](a => a)
+
+  given seqEncoder[A](using innerEncoder: Encoder[A]): Encoder[Seq[A]] = iterableEncoder.contramap[Seq[A]](a => a)
+
+  given indexedSeqEncoder[A](using innerEncoder: Encoder[A]): Encoder[IndexedSeq[A]] = iterableEncoder.contramap[IndexedSeq[A]](a => a)
+
+  given setEncoder[A](using innerEncoder: Encoder[A]): Encoder[Set[A]] = iterableEncoder.contramap[Set[A]](a => a)
```
lbialy, line 1095: This instance is broken for any situation where an element of the set has the same encoded representation because in `interableEncoder` when you `map` on `iterable` that iterable is a `Set` underneath. It's a gotcha with Sets being subtypes of iterables that have a `map` function because for sets the size of resulting collection is between 1 and n:
```
scala> def test(it: Iterable[Int]): Iterable[String] = it.map(_ => "oops")
def test(it: Iterable[Int]): Iterable[String]

scala> test(Set(1, 2, 3)).size
val res0: Int = 1
```
To make it safe you should sequence on a list in `iterableEncoder` so `Result.sequence(iterable.toList.map(...`

core/src/test/scala/besom/internal/DecoderTest.scala
```diff
@@ -138,20 +139,20 @@ class DecoderTest extends munit.FunSuite:
         then
           Map(
             "foo" -> 10.asValue.asOutputValue(isSecret = false, Nil),
-            "bar" -> List("qwerty".asValue).asValue.asOutputValue(isSecret = false, Nil),
+            "bar" -> Iterable("qwerty".asValue).asValue.asOutputValue(isSecret = false, Nil),
```
lbialy, line 142: are these changes in tests necessary? Lists are <: Iterable

core/src/test/scala/besom/internal/ReSerializerTest.scala
```diff
@@ -16,14 +17,14 @@ class ReSerializerTest extends munit.FunSuite:
...
     assertEquals(reSerialize(true).unsafeRunSync(), true)
     assertEquals(reSerialize(false).unsafeRunSync(), false)
-    assertEquals(reSerialize(List.empty[String]).unsafeRunSync(), List.empty[String])
-    assertEquals(reSerialize(List("asdf", "qwer")).unsafeRunSync(), List("asdf", "qwer"))
-    assertEquals(reSerialize(List(1, 2, 3)).unsafeRunSync(), List(1, 2, 3))
-    assertEquals(reSerialize(List(1.0, 2.0, 3.0)).unsafeRunSync(), List(1.0, 2.0, 3.0))
+    assertEquals(reSerialize(Iterable.empty[String]).unsafeRunSync(), List.empty[String])
+    assertEquals(reSerialize(Iterable("asdf", "qwer")).unsafeRunSync(), List("asdf", "qwer"))
+    assertEquals(reSerialize(Iterable(1, 2, 3)).unsafeRunSync(), List(1, 2, 3))
+    assertEquals(reSerialize(Iterable(1.0, 2.0, 3.0)).unsafeRunSync(), List(1.0, 2.0, 3.0))
```
lbialy, line 23: these tests are quite funny because they are green only due to iterable apply returning a list from what I recall. I guess you should change the rhs too

lbialy (changes requested):
Kinda looks like you just went with mass replace but no, you did not. Anyway, some fixes necessary.

core/src/test/scala/besom/internal/DecoderTest.scala
PawelGizka: Nope. These changes were necessary before I created encoders and decoders for list. (iterable decoder and encoder was not enough).

core/src/test/scala/besom/internal/ReSerializerTest.scala
PawelGizka: same here. I also reverted changes in EncoderTest and ResourceDecoderTest

lbialy (approved):

Issue #370, opened 2024-02-06
Add Kubernetes provider overlays


pawelprazak:
We are missing hand-crafted 'overlays' in Kubernetes provider:
- [ ] CustomResource, CustomResourcePatch - frequently used k8s API (top priority)
- [ ] ConfigFile, ConfigGroup, Invokes, TransformationAction - yaml tools (2nd priority)
- [ ] `helm` abstractions (nice to have)
- [ ] `kustomize` abstractions (nice to have)

---
Upstream:
https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/gen/overlays.go
https://github.com/pulumi/pulumi-kubernetes/tree/master/provider/pkg/gen/nodejs-templates

pawelprazak:
IIUC once [pulumi/#1871](https://github.com/pulumi/pulumi-kubernetes/issues/1971) is done we won't have to add a hand-written overlay

pawelprazak:
the simplest solution would be to add `codegen/resources/overlays` and re-use `ProviderConfig` and reuse the existing codegen

pawelprazak:
Those new changes look relevant:
https://github.com/pulumi/pulumi/pull/15490/files#diff-5eea65419cabefe8dae546369e07a0cdc8d2ed74b58326724b49fba56637f910

pawelprazak:
The helm and yaml support can be considered done via 4.11.0+ provider support.

lbialy:
can we close this now, given this was fixed in upstream and inherited via code generation on our side?

pawelprazak:
We're still waiting for full support (as indicated by checkboxes in this issue): 
- CRDs - https://github.com/pulumi/pulumi-kubernetes/issues/2787
- full helm support (v4) - https://github.com/pulumi/pulumi-kubernetes/issues/2847
- kustomize support (v2) - https://github.com/pulumi/pulumi-kubernetes/issues/2786

More info in the upstream epic: https://github.com/pulumi/pulumi-kubernetes/issues/1971

What we have now is full `yaml/v2`.

mjeffryes:
glad the work to create component versions of these overlays is proving useful to you all. FWIW, we are considering not going ahead with an MLC for CustomResource. Since the only function of this overlay is to pack arbitrary fields from the SDK into a single field for Pulumi, there's not really an equivalent component we could write. Plus the implementations in each SDK are relatively simple, making this overlay reasonable to write and support.

[...]

PR #531, opened 2024-07-08
Add custom resource and custom resource patch overlays

polkx:
Add custom resource and custom resource patch overlays based on:
https://github.com/VirtusLab/besom/issues/370


lbialy (approved):

adriaanm:
This is great! Is it available for use somehow? I'm currently facing issues with CustomResources where I need `deleteBeforeReplace` behavior (Gloo VirtualService), but that's not available for a `ConfigGroup`, which is the only way I could find to use custom resources in besom 0.3.2?

PR #544, opened 2024-09-25
Pass plugin discoverer output via a temp file

prolativ:
Build tools might add some garbage to stdout while running the plugin discoverer.
Printing the result's json into a file should solve this problem and make plugins discovery more reliable.

Also bumping pulumi version used in CI


lbialy (approved):
LGTM

PR #557, opened 2025-03-03
Add readme for publish local

kasiaMarek:


yaga/README.md
```diff
@@ -0,0 +1,17 @@
+## Publish Local
+
+### Prerequisites:
+  - [scala-cli](https://scala-cli.virtuslab.org) - enable power mode by running `scala-cli config power true`
+  - [gh](https://cli.github.com) - log in (running `gh auth token` should return a valid github account token)
+  - [polumni](https://www.pulumi.com/docs/iac/download-install/)
```
prolativ, line 6:
```suggestion
  - [pulumi](https://www.pulumi.com/docs/iac/download-install/)
```

yaga/README.md
```diff
@@ -0,0 +1,17 @@
...
+  - [go](https://go.dev)
```
prolativ, line 7: I don't think yaga requires go by itself - it's needed for besom. BTW we should mention building besom as a prerequisite

prolativ: Or is it about installing just (which probably also should be mentioned here)

kasiaMarek: > BTW we should mention building besom as a prerequisite

It's part of building `yaga` (lower).

> I don't think yaga requires go by itself

`go` it is needed to run `besom` tests, which are ran when you build `besom`.

prolativ (approved):

PR #623, opened 2026-08-18
Hotfix diagnostics: validate per feature, merge matching ranges

lbialy:
Three related problems, all of the "the overlay quietly stopped working" kind, found while triaging `just cli packages generate-all` failures.

1. A typeRename targeting a *type* was reported as skipped even though it applied, and a rename targeting nothing was applied to nothing in silence. Features are now validated separately, with provenance.

2. Only the first matching version-range directory was loaded, so overlapping ranges dropped hotfixes silently. All matching ranges now merge; conflicting tokens fail loudly.

3. azure-native's hotfixes are reorganised accordingly — chaos/* keeps an  upper bound as a record of past workarounds, StorageAutogrow moves to an open-ended range so it keeps applying on 3.x.

Details:

A hotfix file declares two independent features with different targets: fieldRemovals only apply to resources, while typeRename applies to resources *and* types, because PulumiPackageInfo applies renames when parsing both. applyToPackage checked only pulumiPackage.resources and, on a miss, reported the whole entry as skipped — while typeRenames was collected a few lines below with no existence check at all.

That produced both a false alarm and a silent gap. azure-native's StorageAutogrow rename is keyed on an enum type, so generation logged "Resource ... not found in package, skipping hotfix" even though the rename had applied; conversely a rename with a stale or mistyped token landed in typeRenames and renamed nothing, forever, without a word.

Validate the two features separately against both resources and types, and report the outcome per hotfix file. Hotfixes now carry the path they were loaded from, so an obsolete entry names the file to delete or re-range instead of leaving the reader to grep for it. Tokens are still matched exactly: this mechanism exists to disambiguate names that differ only by case (azure-native has both StorageAutoGrow and StorageAutogrow in one module), so case-insensitive lookup would defeat its purpose.

loadPackageHotfixes also took only the first matching version range directory, so overlapping ranges silently dropped hotfixes. All matching ranges now contribute, sorted for determinism, with the same token hotfixed from two ranges failing loudly rather than letting one win at random. Provider hotfixes are removals, so they compose.

This makes overlapping ranges safe to keep, which is what the azure-native reorganisation relies on: the chaos/* field removals stay bounded to 2.89.0:3.8.0 as the record of what those versions needed, while the still-live StorageAutogrow rename moves to ^2.89.0. Note that ^ here parses to FromVersion (>=) with no upper bound, not npm-style caret, so ^2.89.0 covers 3.x. azure-native 3.26.0 now generates 51,855 files and reports "1 applied, 0 obsolete".

Add Config.strictHotfixes (default false) so obsolete hotfixes warn for users but can fail CI, keeping overlays from quietly ceasing to apply as upstream schemas evolve.


lbialy:
I am the Law!

